PR #6, opened 2020-09-01
Google Ads

Geometrically:
Integration with Google Ads


Geometrically:
Going to merge this for now, just to test the system. I will explore other options like EthicalAds

PR #8, opened 2020-09-14
Make responsive

stairman06:
Self explanatory.

### DIFF
--- a/assets/styles/global.scss
+++ b/assets/styles/global.scss
@@ -30,11 +30,6 @@
     * Defaults
    */
   font-size: 18px;
-
-  /**
-    Media Tag Values
-  */
-  --mobile-max-width: 1000px;
 }
 
 body {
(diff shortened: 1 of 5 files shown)


falseresync (approved):
Code looks good, although I would greatly advice to mobile-first approach (min-width, instead of max-width) for maintainability reasons. Using desktop-first is good for sketching, while mobile-first is good for sustaining. I would also advice to move the page label (E.g. “Mods”) to the top bar, next to the logo, when the sidebar menu is collapsed. You may just duplicate the label and show the appropriate one on demand.

stairman06:
Changed to be mobile first, and merged upstream so no more merge conflicts. I didn't add the label to the navbar as I'm unfamiliar with Nuxt and not sure of the proper way to do it.

Geometrically:
Pushing my local changes in an hour, there might be some more conflicts - once those are fixed, this should be good for merge.

stairman06:
Merge conflicts have been resolved and the version selector works well.

Geometrically (approved):

PR #28, opened 2020-11-06
Sort mod versions by publish date

FlashyReese:
It's not actually comparing to sort since the API already returns time in order.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,5 @@
 {
   "semi": false,
-  "singleQuote": true
+  "singleQuote": true,
+  "endOfLine":"auto"
 }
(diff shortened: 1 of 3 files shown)


falseresync:
Is this tested? May you also implement a switch to reverse order?
Code notice: better do the operation once where it fetches the list

FlashyReese:
> Is this tested? May you also implement a switch to reverse order?
> Code notice: better do the operation once where it fetches the list

Yes I have tested this, I'm not good at frontend but I can try implementing a switch.

falseresync:
Just slam it somewhere, design is relatively easy to fix afterwards

falseresync:
Already fixed

PR #34, opened 2020-11-16
[WIP] Rework design

Prospector:


diabolical17 (commented):
Why use css variables in scss code when you can use scss variables? 

```scss
$variableName: #ffffff;
h1 {
    background-color: var($variableName)
}
```

Its better to use scss as css variables are chonky thicc:

```css

:root {
    --variable-name: #ffffff
}

h1 {
    background-color: --variable-name
}

falseresync:
@diabolical17 Because with CSS variables you can make automatic theme switch according to browser settings. For reference see https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

Another reason is that SCSS variables are more difficult to work with in a distributed components system like Vue/Nuxt, because you'd need to include variable everywhere - a suboptimal decision at best

PR #53, opened 2020-12-09
Accessibility Changes

Bad-At-Usernames:
I've fixed a few issues from #50, as well as some problems with icons.


Geometrically (approved):
LGTM. Will wait for false to take a look before merged. Thanks for the contribution!

falseresync (approved):

PR #56, opened 2020-12-15
Mod authors don't give rights to modrinth forever...

Gamebuster19901:
Modrinth's terms has an issue similar to what overwolf used to have:

https://curseforge-ideas.overwolf.com/ideas/CF-I-222

Modrinth does not need a perpetual license to modify or prepare derivative works of Mods. Modrinth should be allowed to retain, but not display, distribute, or perform, or serve copies of mods that have been removed or deleted.


Gamebuster19901:
Converted to draft because this is a legal change, and only the owner should probably be merging this.

Geometrically:
Thank you! This change seems fine for me. If overwolf's lawyers have revised it, it should be safe for us.

PR #64, opened 2020-12-28
Teams

### LINKED ISSUE #61: No <h1> or <h2>, etc. are used in mod pages leading to indexing issues.
Issue is, it's not even used for the mod name, so in [LambdaBetterGrass](https://modrinth.com/mod/lambdabettergrass) I had multiple `<h2>` because `##` of markdown for my description, and as there's no other higher-level headers search engines will take this as the main keyword/point of interest and leads to situation like this:
image

Really not great in my opinion :P

Geometrically:
This PR adds teams into the frontend, which allows users to invite other users to mods. 
Depends on: https://github.com/modrinth/labrinth/pull/120

It also (should) fix #61

### DIFF
--- a/components/DashboardPage.vue
+++ b/components/DashboardPage.vue
@@ -7,6 +7,10 @@
             <ModIcon />
             My mods
           </nuxt-link>
+          <nuxt-link :to="'/dashboard/notifications'" class="tab last">
+            <NotificationsIcon />
+            Notifications
+          </nuxt-link>
           <nuxt-link
             v-if="
               $auth.user.role === 'admin' || $auth.user.role === 'moderator'
@@ -37,6 +41,7 @@
 import ModIcon from '~/assets/images/sidebar/mod.svg?inline'
 import ModerationIcon from '~/assets/images/sidebar/admin.svg?inline'
 import SettingsIcon from '~/assets/images/sidebar/settings.svg?inline'
+import NotificationsIcon from '~/assets/images/sidebar/notifications.svg?inline'
 
 export default {
   name: 'DashboardPage',
@@ -44,6 +49,7 @@ export default {
     ModIcon,
     ModerationIcon,
     SettingsIcon,
+    NotificationsIcon,
   },
 }
 </script>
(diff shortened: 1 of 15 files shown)


Mysterious-Dev (approved):

components/DashboardPage.vue
```diff
@@ -7,6 +7,10 @@
             <ModIcon />
             My mods
           </nuxt-link>
+          <nuxt-link :to="'/dashboard/notifications'" class="tab last">
+            <NotificationsIcon />
```
diabolical17, line 11: Wouldn't it be plausible to add a notifications link on the User drop-down on the navbar along side the dashboard?

diabolical17 (changes requested):
Just one little thing I think would be useful.

components/ModPage.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div class="page-container">
+    <h1 class="hidden">{{ mod.title }}</h1>
```
LambdAurora, line 3: Why hide a `<h1>` instead of just replacing the `<h2 class="title">{{ mod.title }}</h2>` a bit later with a `h1`?

diabolical17: Or just set it to a h1 now and set the font size to a h2?

LambdAurora: The title class already manages the font size from what I see.
So yeah, using a h1 instead of a h2 on line 18 makes more sense :P

components/DashboardPage.vue
falseresync: Add notifications to the dropdown; it is a good place for it

falseresync (changes requested):

components/ModPage.vue
falseresync: Replacing an element when it can be done with CSS is a bad idea and not good for performance and UX

LambdAurora: I'm not sure to understand what you mean by this @falseresync

diabolical17: But font size won't cause the user's web browser to consume 900gb of ram will it?

LambdAurora: @fasleresync the reason was because the `h2` was messing up search engines, and I believe that just putting a hidden `h1` is a bad idea and not good for UX.

Those elements have a meaning outside of style.

LambdAurora (approved):

diabolical17 (approved):

PR #69, opened 2021-01-01
Fix Title for Dashboard pages

Mysterious-Dev:
This PR corrects the title of the dashboard pages for more consistency.

If ever you want to change the title for the mods page, there is no problem.


diabolical17:
This is fixed in an upcoming PR I'm about to make. No need for this now.

PR #84, opened 2021-01-19
Fix summary tags being always escaped by XSS

joaoh1:
This PR adds `summary` tags to XSS's whitelist. Previously, they were escaped, meaning that dropdowns could only have "Details" as the title. Here's an example of it working:
image

### DIFF
--- a/plugins/compiled-markdown-directive.js
+++ b/plugins/compiled-markdown-directive.js
@@ -2,10 +2,17 @@ import Vue from 'vue'
 import xss from 'xss'
 import marked from 'marked'
 
+const options = {
+  whiteList: {
+    ...xss.whiteList,
+    summary: [],
+  },
+}
+
+const configuredXss = new xss.FilterXSS(options)
+
 function compileMarkdown(target, markdown) {
-  target.innerHTML = xss(marked(markdown), {
-    whiteList: Object.assign(xss.whiteList, { summary: [] }),
-  })
+  target.innerHTML = configuredXss.process(marked(markdown))
 }
 
 Vue.directive('compiled-markdown', {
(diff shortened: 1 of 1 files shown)


falseresync:
Are you sure this doesn’t open up bodies for XSS? Have you tested XSSing it?

diabolical17:
From what i read, using whitelist opens bodies up.

joaoh1:
> 
> 
> Are you sure this doesn’t open up bodies for XSS? Have you tested XSSing it?

As far as I know, this should just be the same as before but with summary being whitelisted too. I've double-checked right now and it does indeed escape anything else:
firefox_2021-01-20_22-03-10
image

falseresync:
@joaoh1 seems fine then, except for this:

plugins/compiled-markdown-directive.js
```diff
@@ -3,7 +3,9 @@ import xss from 'xss'
 import marked from 'marked'
 
 function compileMarkdown(target, markdown) {
-  target.innerHTML = xss(marked(markdown))
+  target.innerHTML = xss(marked(markdown), {
+    whiteList: Object.assign(xss.whiteList, { summary: [] }),
+  })
```
falseresync, line 8: I'd do it like this:
```js

const options = {
    whiteList: {
        summary: [],
    },
};
const configuredXss = new xss.FilterXSS(options);

function compileMarkdown(target, markdown) {
  target.innerHTML = configuredXss.process((marked(markdown))
}
```

falseresync (changes requested):

plugins/compiled-markdown-directive.js
joaoh1: Seems good, however, it only missed something to also have the original whitelist's values. I figured out a better way to do that without Object.assign

falseresync (approved):

Geometrically (approved):

PR #115, opened 2021-03-14
Implement header ids

### LINKED ISSUE #112: No id tags added to markdown headers.
Having a quick look at Marked. It seems like this is [configurable](https://marked.js.org/using_advanced#options) using the "headerIds" boolean.
It is worth noting that github for example prefixes all header ids. I'm not sure as to the exact reason of this, it might be to prevent conflicts with ids they're using. It might also be to prevent javascript that's calling `document.getElementById` from getting the user's header instead. I'm not sure if this is required for Modrinth, if so it'll probably require some javascript to get working.

TheEpicBlock:
This pr allows the `id` through the xss filter but prefixes them with `user-generated-`. A script has been added to convert the location hash in the url to the correct id in the page. It also handles repeatedly pressing a link to an anchor.

fixes #112

### DIFF
--- a/plugins/compiled-markdown-directive.js
+++ b/plugins/compiled-markdown-directive.js
@@ -6,18 +6,45 @@ const options = {
   whiteList: {
     ...xss.whiteList,
     summary: [],
+    h1: ['id'],
+    h2: ['id'],
+    h3: ['id'],
+    h4: ['id'],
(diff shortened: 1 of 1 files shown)


Geometrically (approved):

Geometrically:
Thank you for the PR!

PR #123, opened 2021-03-28
Follow page, and edit button consistency

### LINKED ISSUE #122: Add support for follow page in user's panel
Modrinth allows for following nex mods using a button embedded in a mod page, as shown in this screenshot:
[Screenshot of follow feature](https://s3.fr-par.scw.cloud/screenshots-red/ShareX/2021/03/firefox_LQSj94RQHq.png)

But currently does not provide any place to see all followed mods in a single page.
To resolve this issue, we should add a new subpage for followed mods in the same pattern than the profile page, located in the dashboard, with the only difference being the navigation instead of the profile badge.

The user journey would be the following:
Main page - Dashboard - Followed mods

Redblueflame:
This PR has for objective to create a page where you can see every mod you have followed, and add consistency to the placement of buttons.

Here is the current way the setting button is positioned
image

Here is how it looks with the modifications in this branch:
image

This PR also adds a follow panel, available when clicking on Dashboard, then "Followed Mods", where you can access the mod pages, and unsubscribe to them.

This PR still needs some things to be polished:
- [ ] Find a way to display that you are unsubscribing to the mod instead of subscribing

Closes #122


Redblueflame:
Update: Changed the looks of the unfollow page, here is the two variations: 
Solution 1:
image

Solution 2:
(No mouseover)
image
(Mousehover)
image

Geometrically:
I think you should use Solution 2, but use the standard button mouseover color. It should already be set

Redblueflame:
In a user experience standpoint, having a red thing on mouseover effect explains that this action is not without consequences, and after testing, having the normal mouseover just didn't bring the effect needed

Geometrically:
https://falseresync.ru/blog/modrinth-knossos-design-guidelines I would refer here, but I think the action isn't severe enough to have a bright color. Currently, no button on the site has a red/ color that pops out. If it is a color, it tends to be the Modrinth green. 

Also, the buttons are taking up too much space in the flexbox, so the description is getting squashed. Could you fix that?

Redblueflame:
This PR seems ready to merge, @Geometrically could you do a final check on the changes?

Geometrically (approved):

PR #126, opened 2021-03-28
More secure custom authentication solution

### LINKED ISSUE #117: Auth Tokens are stored insecurely
They're in localstorage *and* session-only cookies without the proper headers set.
Apparently, this is the default for nuxt/auth.

They *should* be in `Secure`, `HttpOnly`, `SameSite=Strict` cookies only, to soften the blow of a potential XSS.

Geometrically:
Fixes #117

### DIFF
--- a/middleware/auth.js
+++ b/middleware/auth.js
@@ -53,8 +53,8 @@ function routeOption(route, key, value) {
 function getMatchedComponents(route, matches) {
   return [].concat(
     ...[],
-    ...route.matched.map(function (m, index) {
-      return Object.keys(m.components).map(function (key) {
+    ...route.matched.map((m, index) => {
+      return Object.keys(m.components).map((key) => {
         matches.push(index)
         return m.components[key]
       })
--- a/middleware/auth.js
+++ b/middleware/auth.js
@@ -1,0 +1,61 @@
+export default async function (context) {
+  if (!context.from) {
+    if (context.route.query.code) {
+      context.app.$cookies.set('auth-token', context.route.query.code, {
+        secure: true,
+        sameSite: 'Strict',
+        httpOnly: true,
+      })
+
+      return context.redirect(context.route.path)
+    }
+
+    if (context.app.$cookies.get('auth-token')) {
+      const cookie = context.app.$cookies.get('auth-token')
+
+      await context.store.dispatch('auth/fetchUser', { token: cookie })
+    }
+  }
+
+  if (routeOption(context.route, 'auth', false)) return
+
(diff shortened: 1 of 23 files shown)


Redblueflame (approved):
LGTM

Redblueflame (commented):
Just a note, if vue-persist is activated on the store, the auth store will also be stored in the localstorage.

Geometrically:
We don't use vue-persist yet, but it may be added in the future to assist with caching. I believe you can exclude stuff from vue persist though

Redblueflame:
Then everything is fine :)

middleware/auth.js
```diff
@@ -0,0 +1,61 @@
...
+        )
+      )
+    }
+  })
+}
+
+function getMatchedComponents(route, matches) {
+  return [].concat(
+    ...[],
+    ...route.matched.map(function (m, index) {
```
falseresync, line 54: Why not arrow functions? ie `(param) => result`

middleware/auth.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    if (context.app.$cookies.get('auth-token')) {
+      const cookie = context.app.$cookies.get('auth-token')
+
+      await context.store.dispatch('auth/fetchUser', { token: cookie })
+    }
+  }
+
+  if (routeOption(context.route, 'auth', false)) return
```
falseresync, line 20: What are these supposed to check? They aren’t self documenting enough, some commentary is required

plugins/auth.js
```diff
@@ -0,0 +1,3 @@
+export default ({ store }, inject) => {
```
falseresync, line 1: What for? Stores are always available. And there are Vuex store “getters” for easier access

falseresync (changes requested):

plugins/auth.js
Geometrically: The plugin is just for a utility to make it shorter to access the auth stuff. What do you mean by Vuex store getters?

middleware/auth.js
Geometrically: This is taken from: https://github.com/nuxt-community/auth-module/blob/dev/src/core/middleware.ts#L6

middleware/auth.js
Geometrically: Could you suggest what it should look like?

Scotsguy (approved):
Secure, SameSite=Strict, HttpOnly cookies, LGTM. Can't comment on the js quality

falseresync (approved):
Lgtm

PR #128, opened 2021-03-29
Add version support for the docker builds & local builds

Redblueflame:
Add experimental support for github actions


Redblueflame:
Here is how it looks like as it is currently:
image

rubybb (approved):
looks good imo

falseresync (approved):
Design part is tidy

PR #135, opened 2021-03-30
Confirm popups

Redblueflame:


Geometrically (approved):

components/ConfirmPopup.vue
```diff
@@ -0,0 +1,152 @@
...
+          <span class="confirmation-text">{{ confirmationText }}</span>
+          to continue
+        </span>
+      </label>
+      <input
+        v-if="hasToType"
+        id="confirmation"
+        v-model="confirmation_typed"
+        type="text"
+        placeholder="Type your username to continue"
```
rubybb, line 20: Confirmation popup placeholder shouldn't be bound to username?

Redblueflame: Let me fix that so something more generic

Redblueflame: Done

components/ConfirmPopup.vue
```diff
@@ -17,7 +17,7 @@
         id="confirmation"
         v-model="confirmation_typed"
         type="text"
-        placeholder="Type your username to continue"
+        placeholder="Type the input needed to continue"
```
rubybb, line 20: seems strangely worded, but that works I suppose.

Redblueflame: If you have a better suggestion, please tell it, because I was stuck for a good minute trying to find a better wording

rubybb: I thought about it for a bit but couldn't come up with anything better aside from maybe exposing a prop for the component to have it defined on a case by case basis.

Redblueflame: I prefer to avoid adding another prop to the mix, it will only add more complexity to the component

PR #155, opened 2021-04-06
Change ads provider, and add consent system for advertising

Redblueflame:
What's left to do:
- [ ] Add consent withdrawal support for analytics (unami)


Redblueflame:
@Geometrically Everything should be ready for review.

modules/analytics/lib/templates/plugin.js
```diff
@@ -0,0 +1,58 @@
...
+    // Rely on the client
+    cookies = document.cookie;
+  }
+  let processed = {}
+  cookies.split(';').forEach((e) => {
+    let val = e.trim().split('=');
+    processed[val[0]] = decodeURI(val[1]);
+  })
+  let scopes = decodeURIComponent(processed['modrinth-scopes']).split(",");
+  return (scopes !== null && scopes.includes('analytics'));
```
Geometrically, line 18: I would use `nuxt-cookies-universal` for this. It's a lot cleaner and the auth already uses it.

pages/dashboard.vue
```diff
@@ -2,7 +2,7 @@
   <div class="page-container">
     <div class="page-contents">
       <div class="sidebar-l">
-        <div class="card page-nav">
+        <div v-if="$auth.user != null" class="card page-nav">
```
Geometrically, line 5: Why not just `$auth.user`

components/ads/Advertisement.vue
```diff
@@ -1,34 +1,36 @@
...
     },
-    pageUrl: {
+    adUnit: {
       type: String,
-      required: false,
-      default: '',
+      required: true,
     },
   },
+  mounted() {},
```
Geometrically, line 23: Why the empty mounted hook?

Geometrically (changes requested):
Looks good! Just a couple minor changes, thank you so much :+1:

modules/analytics/lib/templates/plugin.js
Redblueflame: Unfortunatly, it's in a module, so dependencies are not allowed there

pages/dashboard.vue
Redblueflame: Just to explain what it does, having a inherent false if empty is just too cryptic to be in there

components/ads/Advertisement.vue
Redblueflame: Let me remove that

Geometrically (approved):

PR #156, opened 2021-04-08
Partial Mobile Support

Geometrically:
This PR adds partial mobile support, with some missing pages related to mod management, which aren't really useful for the typical mobile user.

These pages include:
Mod Create/Edit Page
Version Create/Edit Page
Mod Settings Page (Partial Support, but teams management interface is glitchy)

All other pages work and look fine on mobile!


Mysterious-Dev (approved):
It's good in my test.

falseresync (changes requested):
Close navabar on route change
Use `watch` hook for that: https://stackoverflow.com/questions/42855985/vuejs-close-hamburger-menu-on-route-change

components/ui/ProjectCard.vue
```diff
@@ -1,103 +1,109 @@
...
-        <p v-if="author" class="author">
-          by <nuxt-link :to="'/user/' + author">{{ author }}</nuxt-link>
-        </p>
+    <div class="columns">
+      <div class="icon">
+        <nuxt-link v-if="isModrinth" :to="'/mod/' + id">
+          <img
+            :src="
+              iconUrl
+                ? iconUrl
```
rubybb, line 9: Use the ``Logical OR (||)`` operator.

components/ui/ProjectCard.vue
```diff
@@ -1,103 +1,109 @@
...
+        </div>
+        <p class="description">
+          {{ description }}
+        </p>
+        <div :class="{ vertical: editMode }" class="bottom">
+          <div class="stats">
+            <div v-if="status !== null" class="stat">
+              <div class="info">
+                <h4>Status</h4>
+                <span v-if="status === 'approved'" class="badge green">
```
rubybb, line 36: Does vue have switch cases?

falseresync: No

components/ui/ProjectCard.vue
Redblueflame: Or the ES6 ??, more stable and not relying on some js trickery

rubybb: > Or the ES6 ??, more stable and not relying on some js trickery

In this case it'd be better to use ``||`` as values like an empty string should just default to the placeholder icon, with ``??`` it  would not.

PR #183, opened 2021-04-22
Fix "Available For" field who is set to None

Mysterious-Dev:
This PR fix this issue #182 .

image


Mysterious-Dev:
It's good for review, I changed the way for the fix of the issue.

PR #198, opened 2021-05-04
Add right arrow light to fix wonky arrow colours on home page

wafflecoffee:
Also fixes a typo
Reported at https://discord.com/channels/734077874708938864/783153806531100673/837843248675291176

~~Does it work? Who knows! Neither Vercel nor my own local system want to build, but they give different errors. :D~~ builds now


wafflecoffee:
ffs

wafflecoffee:
dark mode
light mode
Wahoo.

Geometrically (changes requested):
This is a bad way of implementing this, you should use the SVG inline component tool thing (forgot what it's called) and just set the style using vue to bind the style

Issue #100, opened 2021-02-03
Client-side/server-side filters


Madis0:
The search function should allow filtering by client-side and server-side options.

PR #204, opened 2021-05-11
Client-side/server-side filters

venashial:
## Result:
image

----

## Toggle explanation:
image

---- 

Closes modrinth/knossos#100


pages/mods.vue
```diff
@@ -504,11 +540,16 @@ export default {
         ]
 
         if (this.query.length > 0) {
+          console.log(this.query)
```
SwanX1, line 543: does this debug line need to be here?

SwanX1 (commented):
I don't believe the two debug `console.log`s should be there, but otherwise everything else looks good.

pages/mods.vue
venashial: Nope, my bad. Should be fixed with 62c3404b8cf8cd78b29f3b20a661ddf992cceb25

SwanX1 (approved):
Looks good to me, there are no lint errors when running `npm run lint`, and works on my local build as expected.

pages/mods.vue
```diff
@@ -583,7 +652,6 @@ export default {
           window.history.replaceState(new Date(), 'Mods', url)
         }
       } catch (err) {
-        // eslint-disable-next-line no-console
```
Redblueflame, line 586: Keep the eslint checker tho, as it could cause issues when having eslint enabled (for example in the CI runs).

Redblueflame (approved):
Thanks for this PR!
There is a small nitpick that could block the CI, but other than that, I'll be happy to merge it :)

pages/mods.vue
venashial: Sorry, I thought that I added those when I added my console.log()s. I added them back in 3d9be6d.

Mysterious-Dev (approved):

pages/mods.vue
```diff
@@ -521,6 +561,34 @@ export default {
             formattedFacets.push(versionFacets)
           }
 
+          if (this.selectedEnvironments.length > 0) {
+            let environmentFacets = []
+            if (
```
falseresync, line 566: This whole thing can be simplified:
```js
let includesClient = this.selectedEnvironments.includes('client');
let includesServer = this.selectedEnvironments.includes('server');
if (includesClient && includesServer) {
    environmentFacets = [
        ['client_side:required'],
        ['server_side:required'],
    ]
} else {
   ...
}
```
This way it is more visually appealing and makes less searches

falseresync (changes requested):

pages/mods.vue
venashial: Will do, thanks.

falseresync (approved):

SwanX1 (approved):

Geometrically:
Thanks for the PR :)

PR #206, opened 2021-05-12
Fix notification & follow list sorting, Add notification badge + loading animation

venashial:
* Fixes modrinth/knossos#195: (Notifications and followed mods are not ordered)

* Closes modrinth/knossos#145: 
image

* Adds a link to settings in the drop down menu

* Adds a loading animation:
ezgif com-gif-maker

-----

Although the loading animation is only shown for a second or less, I've found it improves the user experience.


Mysterious-Dev (approved):

components/ui/search/LogoAnimated.vue
```diff
@@ -0,0 +1,130 @@
+<template>
+  <div>
+    <svg
```
Redblueflame, line 3: As this is in a separate component, and the remain element, there is no need to inline it.

layouts/default.vue
```diff
@@ -215,6 +234,23 @@ export default {
       this.$colorMode.preference =
         this.$colorMode.value === 'dark' ? 'light' : 'dark'
     },
+    async updateNotifCount() {
+      if (
+        this.$auth.user.id &&
+        Date.now() - this.notifications.lastUpdated > 300000
```
Redblueflame, line 240: Is the delay really needed? 
It would be useful in the case where the notification amount is stored in the vuex store and persisted, but it is not the case, and the `created` method is only called once per element, so that auto update isn't working.

pages/dashboard/follows.vue
```diff
@@ -64,11 +64,12 @@ export default {
       data.$auth.headers
     )
 
-    const mods = (
+    let mods = (
```
Redblueflame, line 67: Why not simply chain the calls instead of mutating the variable?

pages/dashboard/notifications.vue
```diff
@@ -60,12 +60,16 @@ export default {
...
   },
   async asyncData(data) {
-    const notifications = (
+    let notifications = (
       await axios.get(
         `https://api.modrinth.com/api/v1/user/${data.$auth.user.id}/notifications`,
         data.$auth.headers
       )
     ).data
+    notifications = notifications.sort(
```
Redblueflame, line 69: Same here, and remove the console.log, it will be causing errors in the future.

Redblueflame (changes requested):
There is some changes left to to, but all in all, it is a great PR :)

components/ui/AvatarIcon.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <div class="avatar-icon">
+    <svg
```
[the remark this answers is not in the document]
venashial: Can I still use Vue template syntax and things like `v-if` when it is inlined as an SVG via Nuxt? I didn't think that was possible. Inside the SVG, it links to the profile picture and switches between a masked group and a non-masked group.

components/ui/search/LogoAnimated.vue
venashial: I'm not sure what you mean. I could use Nuxt to separate each of the SVGs into the image folder. I avoided that because, in reality, the individual SVGs (that make up the outer & inner rings + the wrench) will only be seen together, on top of each other.

Redblueflame: I just wanted to say that this was in a good spot, no need to move it.

layouts/default.vue
venashial: The delay is needed because `this.updateNotifCount()` is called on route changes too. Storing the notification count in the vuex store seemed overkill because the notification count is only needed in `default.vue`.

pages/dashboard/follows.vue
venashial: It just made more sense conceptually to me, but I can chain the calls.

pages/dashboard/notifications.vue
venashial: Good catch, I really need to watch out for leaving console.log()s in my commits 😬

layouts/default.vue
Redblueflame: The store is never overkill, as it allows us to know for sure that this is dynamic and getting insights in the vue debug tools.
But I can't seem to find the place where you call the updateNotifCount.

components/ui/search/LogoAnimated.vue
venashial: Oh, okay

layouts/default.vue
venashial: Its called right here:
https://github.com/venashial/knossos/blob/3c44b43c1f8128758ad92365727132dde5505f86/layouts/default.vue#L195

Do you think it is still worth migrating to the vuex store?

Redblueflame: It is still worth, as it will make it clearer on for the next one working on it :)

venashial: Should I create a new vuex module just for notifications? The auth one already stores `userFollows`.

Redblueflame: Yup, creating modules is not pretty costy on the JS side

venashial: Should I move `userFollows` while I'm at it? I could call the store something general like `user`

Redblueflame: Yup! `user` is perfect, and it will be used when we have our custom auth system

venashial: Turns out `userFollows` isn't used anywhere. Not sure why/how it got into the vuex store.

venashial:
The notification count on the dashboard looks like: 
image

layouts/default.vue
venashial: Everything is migrated to a vuex store in d527982.

Mysterious-Dev (approved):

Redblueflame (approved):
Looks good to me !

wafflecoffee (approved):
:sweat_smile:
image
I did get an error somewhere but I'm not sure where nor how to reproduce it, so I guess it's all good for me! In case you were curious for whatever reason what that error was:
<details><summary>Error</summary>

```
 ERROR  Cannot read property 'id' of null                                             08:55:09

  at Store.fetchNotifications (server.js:8692:29)
  at Array.wrappedActionHandler (node_modules/vuex/dist/vuex.common.js:853:23)
  at Store.dispatch (node_modules/vuex/dist/vuex.common.js:518:15)
  at Store.boundDispatch [as dispatch] (node_modules/vuex/dist/vuex.common.js:408:21)
  at VueComponent.created (server.js:5499:17)
  at invokeWithErrorHandling (node_modules/vue/dist/vue.runtime.common.dev.js:1850:57)
  at callHook (node_modules/vue/dist/vue.runtime.common.dev.js:4207:7)
  at VueComponent.Vue._init (node_modules/vue/dist/vue.runtime.common.dev.js:4989:5)
  at new VueComponent (node_modules/vue/dist/vue.runtime.common.dev.js:5134:12)
  at createComponentInstanceForVnode (node_modules/vue-server-renderer/build.dev.js:8184:10)
```

</details>

components/ui/AvatarIcon.vue
falseresync, line 3: This is horrendous. It can be achieved with DIVs and positioning.

components/ui/AvatarIcon.vue
```diff
@@ -0,0 +1,124 @@
...
+    align-items: center;
+
+    background-color: #e02914;
+    color: white;
+    border-radius: 100%;
+    width: 0.9rem;
+    height: 0.9rem;
+    font-size: 0.6rem;
+
+    &.smaller {
```
falseresync, line 116: If you use DIVs you can get rid of this and badge will just resize on itself (ie circle -> pill)

falseresync (changes requested):
It was a good idea to add a new Vuex submodule. There are some issues though

components/ui/AvatarIcon.vue
venashial: Do you mean the SVG in general? I am using it to mask the image, but you are right, it isn't very intuitive and requires a vector editor to change it. I could just not use an SVG and instead make a border around the red bubble that matches the header color.

components/ui/AvatarIcon.vue
venashial: I couldn't change the size of the badge because the cutout can't be enlarged. I'll just abandon the SVG cutout all together.

layouts/default.vue
```diff
@@ -177,8 +194,14 @@ export default {
     $route() {
       this.$refs.nav.className = 'right-group'
       document.body.style.overflow = 'auto'
+      if (Date.now() - this.$user.notifications.lastUpdated > 300000) {
```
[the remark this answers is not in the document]
venashial: It could be transferred, though this uses the least amount of code. The delay is only needed for client side route changes.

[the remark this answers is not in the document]
venashial: I moved the delay check to the action in d5e4d38.

components/ui/AvatarIcon.vue
venashial: Pill morphication is added in 5fb6d9d.

components/ui/AvatarIcon.vue
venashial: No more SVG with 5fb6d9d, but choosing the correct border color for the notification bubble required some extra CSS and a prop. Though, overall, I do think it is simpler.

venashial:
> I did get an error somewhere but I'm not sure where nor how to reproduce it

@wafflecoffee I think I caught the source of the error, the code that checks if a user ID is present fails if the user object does not exist. Should be all fixed.

PR #210, opened 2021-05-16
Add show snapshots to URL + Add snapshot checkbox

### LINKED ISSUE #142: [Suggestion] Make the "show snapshots" button in the filters panel, a checkbox
Probably by changing the icon depending on the state, and removing the "shade" when on.
Ideas for the icons are: check and cross, and empty-checkbox and checked-checkbox

*Split from #46*
*Supersedes #49*

### LINKED ISSUE #209: "Include snapshots" Toggle Does Not "Retain" Selection
**Describe the bug**
When selecting the toggle to "Include snapshots" and then navigating away from the page, the selection is not remembered when navigating Back.

**To Reproduce**
1. Select the "Include snapshots" toggle
2. Filter by a snapshot of your choice (optional)
3. Click on any mod in the search/filter results to navigate to the mod's respective page (open mod page in the same tab)
4. Navigate back to the search/filter results by using your browser's back function, a back button on your mouse, etc.
5. The "Include snapshots" toggle is no longer selected and additional snapshots cannot be selected from the dropdown.

**Expected behavior**
As all other filters and selections are remembered when navigating to a mod's page and then back to search/filter results, the expected behavior is for all selections to be automatically restored.  When navigating back to the results, "Include snapshots" should be selected and I should be able to select additional snapshots from the dropdown to add to the filter criteria without re-selecting the toggle.

**Additional context**
This behavior was observed on Firefox 88.0.1 on Windows 10.  I have not tested other browsers/operating systems to determine if the results are the same.

venashial:
Snapshot checkbox looks like this:
image

Fixes #209
Closes #142


assets/images/utils/check.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round"
+	stroke-linejoin="round" viewBox="0 0 24 24">
+	<path d="M20 6L9 17l-5-5" />
+</svg>
```
Geometrically, line 4: Optimize the SVG (like you did in the other PR)

components/ui/search/SearchCheckbox.vue
```diff
@@ -0,0 +1,77 @@
...
+
+  svg {
+    color: var(--color-brand-light);
+    stroke-width: 0.2rem;
+    height: 1.2rem;
+    width: 1.2rem;
+    flex-shrink: 0;
+  }
+}
+</style>
```
Geometrically, line 77: Why is this a separate component (and an odd one too). If you want to make it separate, then make a checkbox component which can be used across the site. This is only used once

components/ui/search/SearchFilter.vue
```diff
@@ -45,7 +45,7 @@ export default {
   align-items: center;
   cursor: pointer;
   padding: 0.4rem 0.3rem;
-  margin: 3px 0 0 0.5rem;
+  margin: 3px 0.5rem 0 0.5rem;
```
Geometrically, line 48: what's the point of this?

Geometrically (changes requested):
It's okay. Could you explain why the spacing was changed?

assets/images/utils/check.svg
venashial: I already optimized it, though my editor added the newlines. Some other SVGs also have newlines so I left them in.

components/ui/search/SearchCheckbox.vue
venashial: You are right, it is pretty odd. It's a copy of `SearchFilter.vue` with some modifications. The `showVersions` array in the `mods.vue` file is already odd, it's an array that is treated like a bool. I could make the component work with `v-model` and move all the spacing over to the filters panel to make it more reusable.

components/ui/search/SearchFilter.vue
venashial: It's part of an attempt to make the search filters panel spacing more uniform. A clearer way of doing this would be to move all the spacing CSS to the parent element. Moving the "clear filters" button to the bottom of the panel would also make things prettier.

assets/images/utils/check.svg
Geometrically: You don't need the XMLNS attribute, just a nitpick though (since the svgs are injected)

venashial: I left it in because it's in a fair amount of other SVGs, I can remove it from all of them though.

components/ui/search/SearchFilter.vue
venashial: I removed all the spacing changes in 4c5f116

assets/images/utils/check.svg
venashial: Done in 4c5f116

components/ui/search/SearchCheckbox.vue
venashial: It's now reusable with 4c5f116

venashial:
I made the checkbox component simpler, and it works with v-model, just like a normal checkbox. I replaced the normal checkboxes with it on the team permissions (mod settings) page.

Geometrically (approved):
Thanks for the PR!

PR #214, opened 2021-05-17
Switch ads provider, and switch the analytics system to Ariadne

Redblueflame:


Redblueflame:
Please don't merge without knowing what you do. There is some work needed on the deployment side, as a new variable was set.

package.json
```diff
@@ -19,10 +19,12 @@
     "@nuxtjs/robots": "^2.4.2",
     "@nuxtjs/sitemap": "^2.4.0",
     "@nuxtjs/style-resources": "^1.0.0",
+    "axios": "^0.21.1",
     "cookie-universal-nuxt": "^2.1.4",
     "highlight.js": "^10.3.2",
     "marked": "^2.0.0",
-    "nuxt": "^2.15.5",
+    "nuxt": "^2.14.7",
```
Geometrically, line 26: Why'd you downgrade Nuxt?

Geometrically (changes requested):

package.json
Redblueflame: This may be due to the rebase errors that I had

PR #220, opened 2021-05-20
Fix report button on own uer page

Mysterious-Dev:
This PR remove the report button on your own user page.

Fix #215 

image

PR #222, opened 2021-05-20
Hide report button on user's own page

venashial:
Reimplementation of #220 using fix outlined in #221


venashial:
Uh oh this is not good, way to many commits.

Geometrically (approved):

venashial:
Everything should be fixed.

PR #226, opened 2021-05-25
Small Landing Page Fixes

WickedTree:
Fixes some coloring of things in different color modes.
Better version of #198
Adds webm video format for the demo video, since some browsers don't support mp4, and webm renders "slightly" better.


Geometrically (approved):
Thanks for the PR!

PR #229, opened 2021-05-26
More UI/UX fixes

### LINKED ISSUE #208: Make download buttons stand out and download icon be reserved for actual downloading
I think the download buttons for mods should be a blue color to help make it stand out against all the options on the website. There's so many buttons and stuff but users are going to be mainly looking for the download button. And by having the button blue like links are, it signals that the button is a link too.
Untitled

In this mock-up picture, the blue helps to know what is clickable as in the bottom picture, the downloads count icon tricked me into thinking that's where I download the mod. This is because the button uses the exact same icon is used for downloading in the top picture.

Having an icon be used for two different purposed is generally bad design as now the users has to think about where they are to know if they can click it instead of recognizing the icon as being a download button right away. 

So my thought is, perhaps change the download icon in the bottom picture to a different icon or make it also be a second button to download the mod. If it is a download link too, then it should be colored blue like the regular download button to show that either of these buttons can be used to download the mod. 

I think this small change will greatly help navigability of the site for people unfamiliar with it as the color really helps draw their attention the button they will be looking for 99% of the time.

### LINKED ISSUE #217: ad text does not update colour when changing theme
**Describe the bug**
Ad text does not change colour when the theme is changed, rendering it nearly unreadable.

**To Reproduce**
1. Go to a location with an ad, such as the [hydrogen mod page](https://modrinth.com/mod/hydrogen).
2. Swap theme.
3. See that the ad's text colour has not changed, making it difficult to read.

**Expected behavior**
The ad's text colour should have changed to match the current theme.

**Additional context**
This is a very minor issue, but I figured I'd just report it so it's known.

WickedTree:
Fixes #217 (Haven't tested), and fixes #208
Makes download buttons stand out more

Before:
image
image

After:
image
image


components/ads/Advertisement.vue
```diff
@@ -109,6 +109,7 @@ export default {
     // Register hook on resize
     window.addEventListener('resize', this.handleWindowResize)
     this.isDark = this.$colorMode.value !== 'light'
+    window.addEventListener('darkMode', this.handleDarkModeChange)
```
Redblueflame, line 112: The main reason why the ads weren't reactive to the dark mode change was to avoid broken tracking pixels in the process due to issues with Vue & Ethical Ads. This may need more testing than the rest

components/ads/Advertisement.vue
```diff
@@ -151,6 +152,14 @@ export default {
         }
       }, 300)
     },
+    handleDarkModeChange() {
+      // I found issues in firefox with the store and isDark getting out of sync
+      // setTimeout seemed to fix it
+      window.setTimeout(() => {
+        this.isDark = this.$colorMode.value !== 'light'
+        this.refresh_ad()
```
Redblueflame, line 160: There may need to be a debounce system for ads refreshing. Too frequent ads refresh can hurt viewability rate, but it's not really important for ethicalads

pages/mod/_id/version/_version/index.vue
```diff
@@ -345,6 +345,15 @@ export default {
     }
   }
 
+  .accent-button {
```
Redblueflame, line 348: This may be used in other contexts, could it be moved to the components.scss file?

pages/mod/_id/versions.vue
```diff
@@ -38,7 +38,7 @@
                 version.id
               "
             >
-              {{ version.name ? version.name : version.version_number }}
+              <b>{{ version.name ? version.name : version.version_number }}</b>
```
Redblueflame, line 41: Why use a bold element?

Redblueflame (changes requested):
The pull request looks good, there is just some issues with the ads modifications, but other than that, I'll deploy a test environment tomorrow with the necessary values for ethicalads to test everything

pages/mod/_id/versions.vue
WickedTree: oops, must have been testing things, and forgot to remove it. Will fix.

Geometrically (changes requested):
Thank you for the contribution, but I do not like the accent color. It does not fit into the site's color scheme. For the "featured downloads" section, I don't think there's a reasonable argument for including an accent, as they do not need to stand out. However, for the version file page, You should just use the green modrinth brand color for the download button. It would work fine.

Minenash:
I think the download button should be the same color as other colored buttons, green. The blue doesn't really fit visually with the color scheme.

Also, I don't think the featured downloads should be moved. While the screenshoted mod only has 2, a mod can have a good amount, pushing all the info down. Also with the current order, the info and the start of the authors are always in the same place, with only downloads moving a bit.

Redblueflame:
While I do agree that the blue itself is not great and does not fit with our branding, I do think that we may need more accent colors to allow for more UX headroom. Using the brand main color, while great, doesn't really bring attention to the most important parts.

Also, one system that could be great would to to allow for at most two "starred" versions, and keep the rest to the bottom.

wafflecoffee:
> While the screenshoted mod only has 2, a mod can have a good amount, pushing all the info down.

This could probably be solved by limiting it to be 5 or so; I don't really see any reason to have more than that unless it's a massive mod that maintains every single Fabric and Forge version ever released or something.

Overall, I do agree with moving the featured versions up (not the authors though), as (just my two cents) we should be mostly thinking about the end-user; however, as mentioned, the blue is rather unnatural.

WickedTree:
> For the "featured downloads" section, I don't think there's a reasonable argument for including an accent, as they do not need to stand out.

I feel like the prevailing issue is that there is no primary (and fast) way of downloading a mod currently, apart from the featured version section. Currently the **only** way to immediately download a mod (without going to the versions page) is through the featured versions section, which is why I think it should be a primary button, since the primary action users are going to take on a mod page is downloading the mod.

Minenash:
> Also, one system that could be great would allow for at most two "starred" versions, and keep the rest to the bottom.

Wouldn't that defeat the feature version thing? If it's supposed to be the set important enough to be on the frontpage, then I don't see why one should be "more important" then the other. If it's not important enough, it shouldn't be on the frontpage

> Overall, I do agree with moving the featured versions up (not the authors though)

I don't think the author should be put at the bottom, below both the downloads and info, it would make it unimportant

Minenash:
> Currently the only way to immediately download a mod (without going to the versions page) is through the featured versions section,

To me this is fine, because the featured versions were supposed to be what the users would most likely download, I personally don't like the ambiguity of a single unidentified download button

Redblueflame:
In that case, as statistics are not really the most important information, we could theoretically inverse the order, going to the following:
Starred downloads (Latest version for both loaders)
Members
Stats
Featured downloads

Also, the PR is already great as-is, the main question left is the layout and the `secondary` color (currently blue, let's find another taint or whole different color)

Minenash:
What would probably help the featured version thing is the ability to manually add or remove versions from it. No longer want the beta to show because the full release is out? Remove it. A mc release just not used anymore, remove it.

I don't think there would be a need for separation, if we had the ability to choose (in the gui) which releases are shown

WickedTree:
Maybe a featured/starred release section at a top and a "recent releases" version at the bottom? Just an idea...

Redblueflame:
The main objective here is to reduce the clutter in the frame by a maximum, so having a maximum of 2 or 3 versions quickly downloadable and recognizable will be better, while still allowing us to get the latest builds for older versions of Minecraft. if this system goes through, I'll make sure to add the fields to Labrinth to keep it modifiable.

Minenash:
I don't really see a use for recent releases, the only releases that most people care about are there newest ones per mc version + loader (release and beta). If they do want to see recent ones, there is a versions tab that lists all of them

Redblueflame:
> 
> 
> Maybe a featured/starred release section at a top and a "recent releases" version at the bottom? Just an idea... Maybe move this discussion to discord if its going to continue?

I don't think having this discussion here is an issue, it allow us to keep it tidy and organized

WickedTree:
Maybe also have a primary download button that activates a dropdown with "featured" versions for different versions of mc? and rename the current featured versions to recent releases, and then switch the order back to what it was.

Minenash:
> The main objective here [...]

Yeah, I get what you're saying, but I feel like having 3 separate lists for where to download mods, is a bit much, and possibly confusing to users. If the point of this separation is to have a list of most likely downloaded, then a list of one for every mc version + loader, then how do betas fit in? Under the above system, it sounds like the featured version list would be automatic, not customizable, but leaving old betas in the list isn't good, and with 1.17 coming out, I don't think 3 customizable slots is enough if you want to show betas

Redblueflame:
As I said, this will be customizable with the addition of a custom field in labrinth. But this part won't be high priority

Minenash:
> Maybe also have a primary download button that activates a dropdown with "featured" versions for different versions of mc? and rename the current featured versions to recent releases, and then switch the order back to what it was.

I'd be down for a drop down button. Though I don't think there's a need for a recent list anywhere on the main page

Minenash:
> and the secondary color (currently blue, let's find another taint or whole different color)

While I don't disagree with maybe adding more colors to modrinth's color profile, I think the green in this case is fine, since the only parts of the page that use it, is the tab underline and the sign in button

venashial:
Two things:
1. The "Info" card on mod pages does not need a title, especially not one that is centered.
2. As crucial as members are to a mod, the mod info is more important and should be placed above. A player isn't going to care about who made a mod if the mod doesn't even support the modloader they use.

Geometrically:
I don't like how the featured versions are moved either, and I feel like info doesn't need a title. I think this PR could really be simplified to changing the color on the download version page.

Minenash:
For the title, I think it fits better with a title even if not strictly needed (though it definitely shouldn't be centered).

WickedTree:
I'm going to close this and redo the pr so that it only changes the color of the download button. Layout changes can be done later and maybe discussed in its own issue.

Minenash:
Since that PR would be modifying the download button (and well only that) would it be simple to fix #239 in the process? (Basically just darkening the color when hovered)

WickedTree:
no problem. Will be able to (hopefully) fix tmrw

PR #272, opened 2021-06-24
Allow users to login from search page when it is rendered serverside

### LINKED ISSUE #256: Auth forces a `?` to be placed before the `code=`
**Describe the bug**
When on the mods page, it loads as, for example, `staging.modrinth.com/mods?q=`. Once pressing the login button, this becomes `staging.modrinth.com/mods?q?code=gho_codehere` for a brief second... it forces the `?` behind the `code=` when it should either replace the existing `?q=` or put a `&` before it.

**To Reproduce**
1. Go to the `/mods` page on staging
2. Press the login button
3. Watch the address bar! 
4. See how you're not logged in after the fact

**Expected behavior**
It should either precede or replace the `?q=` entirely, or make it use a `&`, so that you can log in properly.

**Additional context**
Only occurs on staging (on bb8fd550 at time of writing), not on 0.1 (prod) though.
Thanks to woodiertexas for finding this.

venashial:
Removes the query, `?q=` from the login URL.

Closes #256

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -199,7 +199,7 @@ export default {
   },
   computed: {
     authUrl() {
-      return `${this.$axios.defaults.baseURL}auth/init?url=${this.$store.app.$config.utils.domain}${this.$route.fullPath}`
+      return `${this.$axios.defaults.baseURL}auth/init?url=${this.$store.app.$config.utils.domain}${this.$route.path}`
     },
     userUrl() {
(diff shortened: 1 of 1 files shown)


venashial:
Deploy: https://knossos-venashial-4vu6dyqtr-venashial.vercel.app/mods

wafflecoffee (approved):

Geometrically (approved):

PR #274, opened 2021-06-25
Convert issue template to Issue form

Mysterious-Dev:
This PR convert issue template to issue form (new feature of Github).


wafflecoffee:
Personally, I don't *really* see this being necessary. Imo, the form feature seems to be for more of a larger organization where maybe you have people providing insufficient data or something like that, and where you do need some sort of restriction to ensure people are staying within the lines of what's needed/necessary/appropriate. I wouldn't close this if I were you (I'd like to see others' opinions on this), but that's just my two cents.

Mysterious-Dev:
Surely, the goal was for the creation of an issue to be a little more "user friendly" than a simple template.

Mysterious-Dev:
@Redblueflame What do you think of this ?

Mysterious-Dev:
@Geometrically For you this PR is useless or not ?

PR #276, opened 2021-06-28
Change username to name

venashial:
The user's name is displayed instead of their username in:
- The site header
- The profile page
- The mod members section

If the user's name is empty, their username will be used instead.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -58,7 +58,9 @@
                 >
                   <button class="control" @click="toggleDropdown">
                     <div class="avatar">
-                      <span>{{ this.$auth.user.name }}</span>
+                      <span>{{
+                        this.$auth.user.name || this.$auth.user.username
+                      }}</span>
                       <AvatarIcon
                         :notif-count="this.$user.notifications.count"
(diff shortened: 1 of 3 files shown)


venashial:
Deploy: https://knossos-venashial-gvh4m5qwn-venashial.vercel.app/

Redblueflame (approved):

Mysterious-Dev (approved):

venashial:
Deploy: https://knossos-venashial-cnejukev4-venashial.vercel.app/

Names on user profiles weren't working, now they are.

Geometrically:
I'm going to close this, I think it would be better to have a PR removing the name field from the frontend entirely, I don't see the use for it and I think it's less confusing for the user if Usernames are used as identification everywhere. It's probably best to remove it from the backend too, but that would be best done once the new authentication system is done.

Minenash:
I think the reason it was there is the less restrictiveness. Like take the name `Jákob Zabek`, as a username it would have to be `JakobZabek` with no ability for spaces or non ASCII letters. A lot of names have non ASCII letters, and I think they should still have the ability to use that name

PR #286, opened 2021-07-16
Makes mod editing only send changed data

venashial:
Instead of sending everything to the backend when editing a mod, this keeps track of the old mod data and before uploading checks the differences between them. Then it only uploads the changes.

Importantly, this fixes an issue where team members with the permission to change the mod body are denied when using knossos. (I would assume the issue also exists the other way around.)

Everything relating to donation links is commented out.

The preview mod icon is smaller.

### DIFF
--- a/pages/mod/_id/edit.vue
+++ b/pages/mod/_id/edit.vue
@@ -16,7 +16,7 @@
         "
         title="Submit for approval"
         class="button column"
-        :disabled="!this.$nuxt.$loading"
+        :disabled="!$nuxt.$loading"
         @click="saveModReview"
       >
         Submit for approval
@@ -24,7 +24,7 @@
       <button
         title="Save"
         class="brand-button column"
-        :disabled="!this.$nuxt.$loading"
+        :disabled="!$nuxt.$loading"
         @click="saveMod"
       >
         Save
@@ -36,7 +36,14 @@
         <span>
           Be creative. TechCraft v7 won't be searchable and won't be clicked on.
         </span>
-        <input v-model="mod.title" type="text" placeholder="Enter the name" />
+        <input
+          v-model="mod.title"
+          type="text"
+          placeholder="Enter the name"
+          :disabled="
+            (currentMember.permissions & EDIT_DETAILS) !== EDIT_DETAILS
+          "
+        />
       </label>
       <h3>Summary</h3>
       <label>
@@ -47,6 +54,9 @@
           v-model="mod.description"
           type="text"
           placeholder="Enter the summary"
+          :disabled="
+            (currentMember.permissions & EDIT_DETAILS) !== EDIT_DETAILS
+          "
         />
       </label>
       <h3>Categories</h3>
@@ -69,6 +79,9 @@
           :limit="6"
           :hide-selected="true"
           placeholder="Choose categories"
+          :disabled="
+            (currentMember.permissions & EDIT_DETAILS) !== EDIT_DETAILS
+          "
         />
       </label>
       <h3>Vanity URL (slug)</h3>
@@ -81,6 +94,9 @@
           v-model="mod.slug"
           type="text"
           placeholder="Enter the vanity URL's last bit"
+          :disabled="
+            (currentMember.permissions & EDIT_DETAILS) !== EDIT_DETAILS
+          "
         />
       </label>
     </section>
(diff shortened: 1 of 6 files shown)


venashial:
Deploy: https://knossos-venashial-ppobc49wr-venashial.vercel.app

Mysterious-Dev (approved):

pages/mod/_id/edit.vue
```diff
@@ -624,8 +639,9 @@ section.mod-icon {
 
   img {
     align-self: flex-start;
-    max-width: 50%;
+    max-width: 6.08rem;
```
Redblueflame, line 642: Why is a precise number needed?

pages/mod/_id/edit.vue
```diff
@@ -341,46 +345,47 @@ export default {
...
       }
+      */
 
       availableLicenses.sort((a, b) => a.name.localeCompare(b.name))
       return {
-        mod,
-        clientSideType: mod.client_side.charAt(0) + mod.client_side.slice(1),
-        serverSideType: mod.server_side.charAt(0) + mod.server_side.slice(1),
+        savedMod,
+        mod: { ...savedMod },
```
Redblueflame, line 376: This isn't a deep copy?
I think this could cause issues in production usage.

Redblueflame (changes requested):
There are some issues with this PR, and this is so many moving pieces that I prefer to keep it for the bigger v2 test

pages/mod/_id/edit.vue
venashial: That's the size of all project icons.

pages/mod/_id/edit.vue
venashial: It has to be a deep copy to work. `savedMod` and `mod` are kept separate until the user clicks `Save`. Ideally, the `savedMod` would be saved on Labrinth or LocalStorage, so changes persist closing tabs.

Geometrically:
I don't see the point - why not just add a check for the body permission and handle it accordingly through the user interface (for example, disabling the input and removing the parameter)? This seems overkill for a small issue

venashial:
I agree, disabling the field would be a nice touch. Due to how the `PATCH` endpoints work, it doesn't logically make sense to send the title to be updated if it never changed. This PR simplifies requests. I didn't think adding a 23 line file was overkill.

Geometrically:
> I agree, disabling the field would be a nice touch. Due to how the `PATCH` endpoints work, it doesn't logically make sense to send the title to be updated if it never changed. This PR simplifies requests. I didn't think adding a 23 line file was overkill.

I guess I used the wrong word. I didn't mean overkill, but it is kind of inefficient (like the deep copy). Copies itself in JS are pretty expensive (for some reason :(). I don't see a problem with sending that data, it makes the request a tiny bit faster but if the problem is the body permissions then I think the easiest solution should be used

venashial:
I didn't know that using the spread syntax was considered inefficient. After some research, it appears that using spread syntax can be inefficient, but only when used with very large objects and/or nested loops. Spread syntax in this use case doesn't cause any perceivable performance difference.

pages/mod/_id/edit.vue
```diff
@@ -491,9 +502,13 @@ export default {
...
-        await this.$router.replace(
-          `/mod/${this.mod.slug ? this.mod.slug : this.mod.id}`
-        )
+        this.savedMod = this.mod
+
+        this.$notify({
+          group: 'main',
+          title: 'Changes saved',
+          type: 'success',
+        })
```
Geometrically, line 511: What's the point of changing this?

pages/mod/_id/edit.vue
```diff
@@ -449,29 +454,35 @@ export default {
...
+          ...({ license_url: modChanges.license_url } || {}),
+          ...({ discord_url: modChanges.discord_url } || {}),
+          ...({ license_id: modChanges.license_id } || {}),
+          ...({ client_side: modChanges.client_side } || {}),
+          ...({ server_side: modChanges.server_side } || {}),
+          ...({ slug: modChanges.slug } || {}),
+          ...(modChanges.license
+            ? { license: modChanges.license.short } || {}
+            : {}),
+          /*
```
Geometrically, line 477: This is fine, but could you add a visual indicator that the user is not allowed to edit the body and details (ex. disable input)? It was overlooked previously but it would be a nice addition to the PR.

Geometrically (approved):

pages/mod/_id/edit.vue
venashial: It makes more sense to me, that the user should stay on the project editing page rather than be redirected after saving. The user only clicked `Save`, not `Save & view result`.

pages/mod/_id/edit.vue
venashial: Sure

venashial: Done :)

Geometrically (approved):
LGTM! Thanks for the PR

PR #292, opened 2021-07-25
Update theme colors, add OLED theme, update some project creation text.

Prospector:
image

New OLED theme:
image


NebelNidas:
Hmm, I have to say I liked the old shade of green better 🤔

PR #297, opened 2021-08-03
Allow specific styling in markdown (Text colors)

venashial:
This PR allows users to style HTML elements in Markdown using the `style` attribute on HTML elements.

### Example
image

### Allowed CSS properties
```js
'font-size',
'font-family',
'color',
'background-color',
'background',
'font-weight',
'font-style',
'text-decoration',
'text-align',
```

### Reasoning
- Currently, Modrinth's main competitor has colored text support, and Modrinth does not.
- Requested feature on Discord.
- Not revolutionary, anything done with styling can already be accomplished with images.
- Why not a custom markdown element?
  - Coloring through pure Markdown syntax is not standardized, all results on Google suggest using HTML
  - Making a custom Markdown element with the current Markdown parser would be more work than switching parsers

### DIFF
--- a/plugins/compiled-markdown-directive.js
+++ b/plugins/compiled-markdown-directive.js
@@ -13,7 +13,7 @@ const options = {
     h5: ['id'],
     h6: ['id'],
   },
-  onIgnoreTagAttr: (tag, value) => {
+  onIgnoreTagAttr: (name, tag, value) => {
     if (tag === 'style') {
       // Convert CSS styles to an array of arrays of each property and value
       const styles = value
--- a/plugins/compiled-markdown-directive.js
+++ b/plugins/compiled-markdown-directive.js
@@ -13,6 +13,35 @@ const options = {
     h5: ['id'],
     h6: ['id'],
   },
+  onIgnoreTagAttr: (tag, value) => {
+    if (tag === 'style') {
+      // Convert CSS styles to an array of arrays of each property and value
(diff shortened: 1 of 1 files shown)


venashial:
Deploy: https://knossos-venashial-kltugcy3w-venashial.vercel.app/mod/mudmod#user-defined-css-styling

Geometrically:
I'm going to close this, allowing arbitrary colors isn't great and doesn't look nice. While CurseForge does have colored text, in my opinion it makes the site look more cluttered, messes up with the theme, and draws the user away from more important stuff

venashial:
I don't see this as making the site more cluttered or messing with the theme because many modders already use images to get around styling restrictions.

Minenash:
I also disagree

1. This would allow important information to be more noticeable. Should bold be enough? Yes. Is it? No.
2. Like venashial said, I've seen people use images to get around website's limiting style. And (when it is actually harder to read), text is still better than an image accessibility wise.

I would agree that more limited styling would be nice (like, I'm not sure a bg color would be great), and limiting it to specific colors that change depending on theme would be better, but I don't think that would work with custom tags, for the reason venashial said

venashial:
I can change the allowed CSS properties to limit the contrast with the existing website styles. 

For example,
```js
'font-size',
'color',
'font-style',
'text-decoration',
'text-align',
```

PR #301, opened 2021-08-23
Adding support for purpur plugins

Redblueflame:
Fixed lint errors


wafflecoffee (approved):

mbax:
Perhaps it would be more effective to label as Bukkit, the base API, or Paper, the platform with the most users? Purpur's usage is tiny, and I imagine the number of plugins that exclusively support its API additions can be counted on one (maybe two) hands whereas Paper has support in major plugins.

image

wafflecoffee:
@mbax I had this same question, and it was answered in the Discord:
> **Emmaffle**: ah, I thought bukkit plugins were going to have a separate section like modpacks
> **Redblueflame**: They will in the future, but adding support now would be great I think
> **Emmaffle**: And why purpur specifically and not any of the other bukkit derivs? I'd understand if it were bukkit in general, but having only purpur seems a bit strange
> **Redblueflame**: It's the one that was asked, will add more later down the road
> **EEstar**: We asked :smile: and we are not allowed to post the plugins that depend on the Purpur API on spigotmc.org

Redblueflame:
This PR will be closed at the moment.
Support for plugins will be moved to v2, in order to provide a more streamlined experience for everyone.
Having a great product is better than rushing through things as fast as possible, as it would cause more issues than solve.

PR #317, opened 2021-11-07
Clean up. Part 2: Desassify

falseresync:


falseresync:
Let me try to refork this

PR #320, opened 2021-11-13
Apiv2-support fixes

venashial:
### Untouched:
11. Once a file has been added in new version creation, it cannot be removed, so an accidental misclick will cause you to have to restart
12. The popup preventing you from leaving new version creation without confirming is gone
13. Upon creating a new version, it tries taking you to /project/<proj id>/version/<version id> instead of /<proj type>/...
14. On the project edit page and the project edit page exclusively, the license on the side just disappears https://i.imgur.com/IBHdnb3.png
15. Additions, edits, and removals to external links, technical information, or gallery are not shown until the page is refreshed
16. Upload image button on gallery is chunky
17. After editing a version, going to the changelog page has the versions out of order

### Weird things:

- `/search` route does nothing
- You can transfer ownership to someone who hasn't accepted your invite. Then you both become "Owner"s but you can't edit any members including yourself, so you are stuck in the project.
- A version can depend the same project multiple times, including its own version

### Fixed:

3. After creating a report for a version, it takes you to /version/<ver id> instead of /<proj type>/<proj id>/version/<ver id>
6. OLED mode isn't listed as an option in the text next to the theme selector in profile settings
10. Remove member and Transfer ownership buttons in project settings don't work

### Can't reproduce:

5. On the search page, if you have "relevance" selected as your sort type, the other options are cropped weird

### Needs styling (False's new BEM PRs?): 

2. Not a bug, but is there really no way to see gallery items in larger view?
- Multiselect needs general styling
9. Text contrast on popups (e.g. errors and confirming privact settings) is abysmal

### Due to backend notifications failing:

1. On a project page, no matter whether you're following it or not, it'll always say you're not following it
4. Every time that you switch pages, the loading bar up top flashes red, despite nothing necessarily having gone wrong
7. Download and follow count is always 0 on profile settings page
8. Followed projects page always flashes nothing followed before showing what's actually followed
18. Backend: Notifications don't work? I don't have any notifications from creating a version on my own project which I follow

### DIFF
--- a/components/ui/SmartFileInput.vue
+++ b/components/ui/SmartFileInput.vue
@@ -1,5 +1,5 @@
 <template>
-  <button class="iconified-button" disabled>
+  <div class="columns">
     <label class="iconified-button" @drop.prevent="addFile" @dragover.prevent>
       <UploadIcon />
       {{ prompt }}
@@ -10,7 +10,7 @@
         @change="onChange"
       />
     </label>
-  </button>
+  </div>
 </template>
 
 <script>
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -1,6 +1,4 @@
-import {
-  sortRoutes
-} from '@nuxt/utils'
+import { sortRoutes } from '@nuxt/utils'
 
 export default {
   /*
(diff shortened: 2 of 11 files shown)


Geometrically (approved):
Looks great! Thanks for the PR

Geometrically:
My commit should fix 11-17. 1-5 are now fixed in backend commits. The first and third weird things are things I don't plan on fixing, and the second one is now fixed through a UI change and a backend fix.

PR #321, opened 2021-11-14
Clarify that Modrinth is only for Minecraft

### LINKED ISSUE #319: Clarify that Modrinth is specifically for Minecraft mods.
The home page (https://modrinth.com/) doesn't mention anywhere that it's specifically for Minecraft. 

I imagined it being a general purpose mod-hosting website like Nexus Mods, until I actually tried to browse the mods.

MarkusG:
This resolves #319.

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -5,7 +5,7 @@
         <h1 class="typewriter">
           {{ currentText }}<span aria-hidden="true"></span>
         </h1>
-        <h1>modding platform</h1>
+        <h1>Minecraft modding platform</h1>
       </div>
       <div class="right columns">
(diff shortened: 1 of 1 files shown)


wafflecoffee (approved):
It's worth noting that Modrinth will likely be expanding into other games in the distant future. For the time being, though, it's only going to be Minecraft and will likely stay that way for quite a while, so this is probably a good change to get in.

Mysterious-Dev (approved):

falseresync (approved):

Mysterious-Dev:
This PR is maybe outdated with recent changes

PR #341, opened 2022-01-12
Fix various issues reported on the Trello

wafflecoffee:
See 286025c's comments for what exactly this addresses - or just look at the "Fix Pushed" column


Geometrically (approved):
LGTM!

PR #350, opened 2022-02-11
Updated the link to the Modrinth contribution guide

clownless:


wafflecoffee:
Fixed by modrinth/docs#23, thanks for noticing this nonetheless!

PR #391, opened 2022-03-06
Make links on project pages more consistent, fix some selectors showing when they shouldn't

wafflecoffee:
summary of changes

### DIFF
--- a/components/ui/VersionFilterControl.vue
+++ b/components/ui/VersionFilterControl.vue
@@ -32,7 +32,10 @@
       @input="updateVersionFilters()"
     ></Multiselect>
     <Checkbox
-      v-if="getValidVersions().length > 1"
+      v-if="
+        getValidVersions().length > 1 &&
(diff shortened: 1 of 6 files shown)


components/ui/VersionFilterControl.vue
```diff
@@ -27,6 +32,7 @@
       @input="updateVersionFilters()"
     ></Multiselect>
     <Checkbox
+      v-if="getValidVersions().length > 1"
```
Prospector, line 35:
```suggestion
      v-if="
        getValidVersions().length > 1 &&
        getValidVersions().some((v) => v.version_type !== 'release')
      "
```

Prospector (changes requested):

Prospector:
I think the underlines are an improvement to help indicate that they are clickable. In a future PR I think I'll make the entire elements clickable, but the underlines will still be a good indicator to even try clicking them.

Prospector (approved):

PR #395, opened 2022-03-06
Some more cleanup to prepare for 2.1.0 release

Prospector:


components/ui/VersionFilterControl.vue
```diff
@@ -7,20 +7,27 @@
       v-if="getValidLoaders().length > 1"
       v-model="selectedLoaders"
```
wafflecoffee, line 8:
```suggestion
      v-model="selectedLoader"
```

wafflecoffee (changes requested):

components/ui/VersionFilterControl.vue
Prospector: oops, this got lost in the merge

wafflecoffee (approved):

Geometrically (approved):

Issue #380, opened 2022-03-05
Integration with Mod Managers


trash-panda-mc:
### Is your feature request related to a problem? Please describe.

Currently, mod managers must use the backend API. This isn't inconvenient, but it makes it much more difficult for Modrinth to serve advertisements and adds additional complexity on the mod manager's developers in that they must make their own browser for mods.

### Describe the solution you'd like

Using the Modrinth website itself as a mod browser would allow for a much more streamlined process of downloading and installing mods and modpacks. This, of course, requires a separate solution for the actual deployment process, but would allow for ad revenue to be collected, greatly lessen the burden on mod managers to create a mod browser of their own, and allow users to experience the latest features developed for Modrinth independent of the mod manager.
A mod manager can host a websocket to be used by Modrinth as a means of establishing communication. In the case that a mod manager is not available via websocket, such as if a mod manager is installed but is not currently running, Modrinth can attempt to revive or start it via a protocol handler installed by the mod manager. From there, the mod manager can send information related to the client to Modrinth to be used as a baseline for searching for mods and modpacks, such as the Minecraft version and installed/preferred mod loader. Modrinth can implement a checkout-style list of mods selected by the user to be sent to the mod manager when the user is done, or simply tell the mod manager that a user has selected a mod to be installed. With this integration, Modrinth can also display which mods have already been installed and optionally hide them if the user wishes.
For example, a user can, in their mod manager, choose to browse for modpacks or select a profile or instance to begin browsing mods for, the mod manager can open the Modrinth website for them if the user doesn't already have it open, select a modpack or mods from a prefiltered list tailored to their profile or instance that they would like to install, and seamlessly have the mod manager begin to download what they requested, directly from Modrinth and presented with advertisements to support the content creators and Modrinth itself.

### Describe alternatives you've considered

Currently, the only alternative is to use Modrinth's API, which doesn't enable the possibility of advertisements being shown and can lead to a less-than-satisfactory experience if a mod manager decides to roll their own mod browser using the API.

### Additional context

This solution seems somewhat easy to implement and, in my opinion, will definitely lead to a better user experience overall.

LoganDark:
> much more difficult for Modrinth to serve advertisements

Uhh, excuse me, what? Why is this such a central theme of this issue?

wafflecoffee:
[...]

PR #417, opened 2022-03-11
Mod Manager Integration

trash-pandy:
A partial implementation for issue #380.
Currently adds a button next to any downloads that launches a protocol handler intended to be received by an application on the user's computer that is configured to handle it. The protocol handler currently takes the form `modrinth:/add-item/<version-id>`, where `<version-id>` is the version of the item selected to be downloaded. The button can be enabled by adding the `manager-integration` parameter to the page URL, which is intended to be used by the application referenced above.


Geometrically (changes requested):
Nice job on the implementation! I am going to leave this on requested changes for now. I think before merging the spec should be documented @ https://docs.modrinth.com. I'll talk with the rest of the team on what the spec would look like + any additional features it may support

MulverineX:
Maybe have the button always be there but unless the cookie has been set have it open to the modrinth launcher page?

trash-panda-mc:
> Maybe have the button always be there but unless the cookie has been set have it open to the modrinth official launcher page?

Perhaps in future, when Theseus is in a more useable state, but that day is yet to come.

trash-panda-mc:
There were some major merge conflicts that I found were much easier to resolve by resetting my fork to the latest upstream commit and re-adding what I had written.

triphora:
Hey @trash-pandy! Would it be possible for you to recreate this PR on master? We're still interested in getting this merged, but the rewrite is on hold the foreseeable future, so it would be better to add this to the current Nuxt site.

Minenash:
If this gets merged, probably would be good to let launchers know this is planned so they can add it before the button appears (I see the polymc one above, but it's still open, and prism is the new poly so)

Also maybe the link should be a bit different, as add-item sounds weird for stuff like modpacks

triphora:
Closing in favor of #783

Issue #467, opened 2022-05-08
Strip any spaces from project ID inputs.


woodiertexas:
### Is your feature request related to a problem? Please describe.

When copying a project's ID, sometimes spaces will get copied as well and when I go to paste the ID into an input like the dependency input for example, I get an error that the project doesn't exist.

### Describe the solution you'd like

Essentially search for any spaces in project ID inputs and trim them.

This is what I copy when trying to copy the project ID belonging to Quilt Standard Libraries:
`
qvIfYCYJ `
![image](https://user-images.githubusercontent.com/17211100/167286758-15d94159-8f40-4a57-9211-ce4a0ab561e7.png)

Mx772:
When copying, double-click the ID and copy. 
You won't get a return character included. 

If you drag-select you are picking up extra characters. This is the case with any of the fields in `Technical Information` for any mod.
The value is correct in that it doesn't have a return character in it. 
![image](https://user-images.githubusercontent.com/9059161/169432996-3fcaf5a6-a74d-437c-b599-31cc5107daef.png)

Just for reference, if you replace the value of the ID with another value with no spaces/etc. It still happens when you drag-copy.

ubionexd:
It is a really annoying thing so i just fixed it.

wafflecoffee:
Fixed in #484

PR #484, opened 2022-05-21
Project ID removes spaces (Issue #467)

ubionexd:
No more annoying spaces, numbers or special characters.
The way i'we done it is probably not the best, but it works!
Fixed GitHub Issue #467

### DIFF
--- a/pages/_type/_id/version.vue
+++ b/pages/_type/_id/version.vue
@@ -398,6 +398,7 @@
             <input
               v-model="newDependencyId"
               type="text"
+              oninput="this.value=this.value.replace(/[^a-z]/gi,'')"
               :placeholder="`Enter the ${dependencyAddMode} ID...`"
             />
             <Multiselect
--- a/pages/_type/_id/version.vue
+++ b/pages/_type/_id/version.vue
@@ -398,7 +398,7 @@
             <input
               v-model="newDependencyId"
               type="text"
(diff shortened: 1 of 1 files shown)


wafflecoffee:
This field accepts slugs, though, and slugs can have other characters (e.g. fabric-api or lovely_snails).

pages/_type/_id/version.vue
```diff
@@ -398,6 +398,7 @@
             <input
               v-model="newDependencyId"
               type="text"
+              oninput="this.value=this.value.replace(/[^a-z]/gi,'')"
```
Prospector, line 401: Should only remove whitespace

Prospector (changes requested):

ubionexd:
I will change it as soon as i can, but im off pc rn. Its only 10 AM.

pages/_type/_id/version.vue
ubionexd: Pushed a commit to this pr. Should be good now.

wafflecoffee (approved):

pages/_type/_id/version.vue
```diff
@@ -398,6 +398,7 @@
             <input
               v-model="newDependencyId"
               type="text"
+              oninput="this.value = this.value.replace(' ', '')"
```
Prospector, line 401: Should we maybe use .trim() so it gets any whitespace?

Prospector (approved):

PR #487, opened 2022-05-21
Fix #485 by removing default loader from version filters

### LINKED ISSUE #485: Default version filter mismatch
### Environment

Website

### Describe the bug

The default value for the version filter in mod changelog and versions pages doesn't match what is actually being displayed

### Steps To Reproduce

1. Visit https://modrinth.com/mod/ears/changelog
2. See how "fabric" is selected in the dropdown, but *all* versions are shown

### Expected behavior

Seeing only Fabric versions

### Additional context

If you change the filter (e.g. to "forge" and then back to "fabric"), then it works correctly.

wafflecoffee:
It's really not our place to be imposing a default loader on people, especially as we start adding more loaders

### DIFF
--- a/components/ui/VersionFilterControl.vue
+++ b/components/ui/VersionFilterControl.vue
@@ -6,7 +6,9 @@
     <Multiselect
       v-if="getValidLoaders().length > 1"
       v-model="selectedLoader"
-      :options="getValidLoaders()"
+      :options="
+        getValidLoaders().map((x) => x.charAt(0).toUpperCase() + x.slice(1))
+      "
       :multiple="false"
       :searchable="false"
       :show-no-results="false"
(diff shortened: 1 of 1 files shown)


venashial (approved):

Prospector:
The reason for a default was that when you go to a mod's versions or changelog (especially changelog), you don't want to see the files for multiple loaders as it ruins the user experience. I would just fix the bug which is that the default is not applied properly.

wafflecoffee:
I mean, the preferred default is going to be different for everyone, and imposing Fabric on everyone is less than ideal. If you take the idea of the default being Fabric and there's a user who only ever uses Forge, that's going to be really annoying to have to switch all pages to that. If someone goes to the Versions tab and wants to download a Forge version of a mod, they *always* have to change the filters. You might say that this also occurs if there is no default selected -- but with no default selected, there's usually a pattern where Forge and Fabric are alternating, so they can just select the one that they want.

Prospector (approved):
eh ok

PR #508, opened 2022-05-27
Add chs localization and language keys for future version

baka-gourd:
- Added Simplified Chinese

## Problems
- `tags.miscellaneous` is same of `tags.misc`


venashial:
Thanks for your contribution! I'm planning to move translations to a separate repository. If you could create another PR there in a bit, that would be great, if not, I can make a co-authored commit.

venashial:
That repository is now live at https://github.com/modrinth/translations

baka-gourd:
One possible problem: independent localization repo means the difficulty of debugging. Actually, some svelte files need to be changed to localize the content. Do I need to submit pr in two warehouses respectively?

baka-gourd:
If I want to translate [version.svelte](https://github.com/baka-gourd/knossos/blob/sveltekit-rewrite/src/routes/%5Btype=projectType%5D/%5Bidentifier%5D/version/%5Bversion%5D.svelte#L156), I need to change the svelte file

baka-gourd:
@venashial This pr needs to be reopened and I have added localization keys for many of the contents

PR #512, opened 2022-05-28
Add date tooltips

TheEssem:
This adds a tooltip that appears whenever you hover over the Created/Updated text in search results and mod pages. Example:
date

### DIFF
--- a/components/ui/ProjectCard.vue
+++ b/components/ui/ProjectCard.vue
@@ -66,11 +66,21 @@
             class="right-categories"
           />
           <div class="dates">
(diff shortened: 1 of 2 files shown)


Minenash:
Is it in local time, server time, or UTC?

TheEssem:
> Is it in local time, server time, or UTC?

Local time; I tested it on another PC (with the server running on my main PC) with a different time zone set system-wide and the time changed accordingly.

Geometrically (approved):

PR #514, opened 2022-05-29
Dehardcode *everything* for i18n

wafflecoffee:
*except legal pages and moderation page


wafflecoffee:
modrinth/omorphia#2 will need to be merged before this

baka-gourd:
Looks like a subtle overlap with https://github.com/modrinth/knossos/pull/508

PR #518, opened 2022-05-30
External dependencies list

Geometrically:


wafflecoffee:
Does not seem to work. [See here](https://knossos-iwmowbl4w-modrinth.vercel.app/modpack/waffles-modpack/version/2022.5.30), this is supposed to have an external dependency with Roughly Enough Items

wafflecoffee:
Oh wait duh it's not on prod yet, disregard

Geometrically:
> Oh wait duh it's not on prod yet, disregard

Won't work when it is either, because I have to update the deps of old modpacks on the platform

PR #531, opened 2022-06-09
Nuxt Season Finale

### LINKED ISSUE #530: Filtering versions/changelogs by version doesn't work anymore
### Environment

_No response_

### Describe the bug

On the website, the Minecraft version filters do not work unless you select exactly one loader

### Steps To Reproduce

1. Go to a mod's page
2. Click versions
3. Select some versions under "Filter versions"
4. Wait for results

### Expected behavior

Expected results but they won't come until a loader is selected. This is bad because e.g. I don't care if a mod developer marked their individual version as Fabric or Quilt. Or maybe I want to see if those 1.14 versions are for Fabric or Forge...

### Additional context

Related: #454

Prospector:
Here it is everybody. This marks the final round of polish for the Nuxt version as we shift focus to the rewrite. This probably has bugs. Please find them! This PR touches almost every bit of the frontend, and has everything from visual changes, to UI changes, to validation changes, to function changes, to new features, to grammar changes. We got it all.

Closes #530


pages/create/project.vue
```diff
@@ -105,6 +156,7 @@
           alt="preview-image"
         />
         <file-input
+          max-size="262144"
```
wafflecoffee, line 159: This needs to be converted to a SmartFileInput for it to work

wafflecoffee (commented):
Looks good for the most part

MMK21Hub:
Posting this issue with the new project validation here, so that it doesn't get buried in Discord chat:

When adding a version for the first time after attempting to save an invalid project, validation is immediately preformed on the version.

**Steps to reproduce:**
1. <https://knossos-nad265pog-modrinth.vercel.app/create/project>
2. Click "Save as a draft"
    → ✅ Validation is performed on the project as a whole, pointing out that (among other issues) a version needs to be added
3. Click "Add a version"
    → ❌ The version form is validated as soon as I attempt to add a version
4. Cancel adding the version
5. Click "Add a version" again
    → ✅ The version form is not immediately validated

**Expected behaviour**
The version form would only be validated once I click "Add".

layouts/default.vue
```diff
@@ -351,6 +368,19 @@ export default {
       this.$store.dispatch('tag/fetchAllTags'),
       this.$store.dispatch('cosmetics/fetchCosmetics', this.$cookies),
     ])
+    if (
+      (this.$auth.user && this.$auth.user.role === 'moderator') ||
+      this.$auth.user.role === 'admin'
```
John-Paul-R, line 373: nb: Since I've seen this condition a few times already: A helper method might be useful here. flow could be something like `$auth.user?.isStaff()` or `$auth.user?.canModerate()`

components/ui/SmartFileInput.vue
```diff
@@ -36,6 +36,10 @@ export default {
       type: String,
       default: null,
     },
+    maxSize: {
+      type: String,
```
John-Paul-R, line 40: nb, as I think this works regardless:

Is there a reason this is `String` (as opposed to `Number`)?

pages/create/project.vue
```diff
@@ -567,39 +795,85 @@
...
+                  Reset
+                </button>
+              </div>
+            </div>
+          </div>
+        </div>
+        <div class="gallery-items">
+          <div
+            v-for="(item, index) in gallery.filter((it) =>
+              currentGalleryIndex !== -1
```
John-Paul-R, line 837: nit: ternaries with boolean literals can usually be more clearly written as boolean expressions:

```suggestion
currentGalleryIndex === -1 || it !== gallery[currentGalleryIndex]
```

(If I'm too nit-picky here btw, feel free to ignore and/or tell me off!)

John-Paul-R:
Most of this looks good to me. Have a few more garden-variety minor suggestions similar in theme to the ones above, but I'm not trying to spam, so lmk if those are actually wanted.

(Main thing is that there is a good bit of repeated logic - places where we'd benefit from some reusable helper functions. (e.g. for capitalizing the first letter of a string))

Overall, 👍

isXander:
I have an idea for UI of the client-side validation

Rather than having lots of unused space forming a list of issues, instead, disable the submit for review button and on hover of said button, have a tooltip listing all the issues (obviously not the system tooltips)

These checks could now be performed in real-time as you type as no error would be revealed until the user comes to submit for review, noticing that they can't.

venashial (approved):

wafflecoffee:
Going to go ahead and merge this as it's rather urgent. We still need a way to show the max size prompts when using StatelessFileInput instead of SmartFileInput, but we can figure that out later.

PR #537, opened 2022-06-13
Create `StatelessFileInput`, use in version.vue

### LINKED ISSUE #536: Project Version Create/Edit: File Input - All previously-added files are duplicated on `change`
### Environment

Website

### Describe the bug

Project Version Create/Edit: File Input - All previously-added file are duplicated on `change`

### Steps To Reproduce

1. Go to the `versions` tab of any Mod page you have write permissions on.
2. Click "Upload" to begin creating a new version
3. Add a file in the `Files` section (`OriginalFile`)
5. Add another file (`SecondFile`)
6. Observe that the original file has been duplicated. (the list is now `[OriginalFile, OriginalFile, SecondFile]`)
### Expected behavior

- On `change`, only the newly-added file should be added to the files list

#### Root Cause

This happens because the "files" form state is stored in two different locations. Within the `SmartFileInput` component, and in `versions[currentVersionIndex].files` in `pages/create/project.vue`.

When `SmartFileInput::change` fires, `versions[currentVersionIndex].files` has all of the previous values re-added, since the every value inside the `SmartFileInput` gets pushed.

Relevant bit:
https://github.com/modrinth/knossos/blob/2fb62fe8bea48022dbb8769431e8cb462a2bd790/pages/create/project.vue#L464-L492

*edited to reduce extraneous/redundant info*

John-Paul-R:
Resolves #536 

In short, state was being duplicated (in the form and in the component). This component has no internal state. Instead, it lets the caller handle state management (communicating additions via the `change` event).

### DIFF
--- a/components/ui/StatelessFileInput.vue
+++ b/components/ui/StatelessFileInput.vue
@@ -1,0 +1,89 @@
+<template>
+  <div class="columns">
+    <label class="button" @drop.prevent="addFile" @dragover.prevent>
+      <span>
+        <UploadIcon />
+        {{ prompt }}
+      </span>
+      <input
+        type="file"
+        :multiple="multiple"
+        :accept="accept"
+        @change="onChange"
+      />
+    </label>
+  </div>
+</template>
+
+<script>
+import UploadIcon from '~/assets/images/utils/upload.svg?inline'
+
+export default {
+  name: 'StatelessFileInput',
+  components: {
+    UploadIcon,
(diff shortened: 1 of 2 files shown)


John-Paul-R:
### Before (`SmartFileInput`)
SmartFileInput

### After (`StatelessFileInput`)
StatelessFileInput

wafflecoffee (approved):

PR #548, opened 2022-06-21
Show report button when logged out

stairman06:


pages/_type/_id.vue
```diff
@@ -109,10 +109,9 @@
             <span class="value">{{ $dayjs(project.updated).fromNow() }}</span>
           </div>
         </div>
-        <hr v-if="$auth.user" class="card-divider" />
+        <hr class="card-divider" />
```
wafflecoffee, line 112: Could this be done on version pages too?

stairman06: Oops, I missed that. Fixed it.

wafflecoffee (approved):

PR #550, opened 2022-06-25
Add "See all" link next to featured versions

bdsqqq:
A couple of friends and I who started using Modrinth recently didn't realize the "versions" tab existed right away. This PR aims to provide a way for users to easily find and access the "versions" tab from the "featured versions" section.

### DIFF
--- a/pages/_type/_id.vue
+++ b/pages/_type/_id.vue
@@ -421,6 +421,16 @@
               />
             </div>
           </div>
+          <div class="links">
+            <nuxt-link
+              v-if="project.versions.length > 0 || currentMember"
+              :to="`/${project.project_type}/${
+                project.slug ? project.slug : project.id
+              }/versions`"
+            >
+              <span>All versions</span>
+            </nuxt-link>
+          </div>
           <hr class="card-divider" />
         </template>
         <h3 class="card-header">Project members</h3>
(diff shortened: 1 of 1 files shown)


wafflecoffee (commented):
It's a good idea, but it looks quite out place right now.
image
imo, it might be better as something like this:
image
But if you or anyone else has any better ideas feel free to elaborate :)

bdsqqq:
Thanks for the feedback!
I really liked the approach you suggested of putting the link in the section header. I'll update the code for the pull request to reflect it as soon as I have some time :).

Minenash:
>  imo, it might be better as something like this:

that does look good, probably look a bit better if the () weren't apart of the link

wafflecoffee:
Is this still being worked on?

bdsqqq:
Changed the label to "(See all)", and moved it to after the "Featured versions" h3.

I kept the ( ) in the link because since the label is inside a `span class="links"` > `nuxt-link` > `span`, it would render "( see all)" eg: see all with an space after the parenthesis. I chose not to mess with the structure to keep this link consistent with the others I found in this file.

Also, to make the link render in the same line as the "Featured versions" `h3` I added `width: fit-content;` and `display: inline;` to `.card-header` 

On desktop:
image

On mobile: 
image

bdsqqq:
btw, @wafflecoffee sorry for the late response and VERY late update. I put this on my todo list and never got to it.
Hope this is still a welcome update as I think it would improve UX.

wafflecoffee (approved):

PR #555, opened 2022-06-26
Hoverable "joined" fields on user pages

ubionexd:
Implemented #540 and now the cursor does not change to "text" on hoverable "date" fields. Instead, it remains the "default" cursor.
I think it looks cleaner, but I can remove it if the Modrinth Team doesn't like it.


ubionexd:
Why is this got closed?

ubionexd:
Thanks :)

wafflecoffee (approved):

PinkGoosik (approved):

PR #566, opened 2022-07-05
Create config.yml for issue selection

Andre601:
Adds a config.yml to `.github/ISSUE_TEMPLATE/` which allows to show links to useful pages.

I decided to include the GitHub Invite and Modrinth documentation page. If any other pages should be added, let me know.

I also left `blank_issues_enabled` on true. If that should be changed, let me know too.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-enable_blank_issues: true
+blank_issues_enabled: true
 contact_links:
   - name: Discord
     about: Ask questions on our Discord Server.
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+enable_blank_issues: true
+contact_links:
+  - name: Discord
+    about: Ask questions on our Discord Server.
+    url: https://discord.gg/EUHuJHt
+  - name: Documentation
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,8 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Discord
+    about: Ask questions on our Discord Server.
+    url: https://discord.gg/EUHuJHt
+  - name: Documentation
+    about: Useful documentation about Modrinth, its API and how you can contribute.
```
wafflecoffee, line 7: Needs an oxford comma

wafflecoffee (changes requested):

.github/ISSUE_TEMPLATE/config.yml
Gaming32: Here's two examples:
  + "foo, bar and baz"
  + "foo, bar, and baz"
The extra comma in the second is an Oxford comma.

wafflecoffee (approved):
Would also be nice if there was a link to labrinth's issue tracker like "if it's a backend issue or request, go here, this is only for frontend issues"

Andre601:
> Would also be nice if there was a link to labrinth's issue tracker like "if it's a backend issue or request, go here, this is only for frontend issues"

Main issue I see here is, that most people who report issues may not even know what the difference between frontend and backend is.
(Also, with the feature of transferring issues between repositories should this be a minor issue?)

PR #578, opened 2022-07-24
Enhance version creation

### LINKED ISSUE #535: Assume version name is the version itself

I name my versions the same as the version itself. I don't want to type that twice...

- Allow me to type only the version name, assume that the version is identical
- Allow me to type only the version, assume that the version name is identical

A button to copy value from the version?

### Additional context

_No response_

kevoh1516:
My first PR go easy on me. Initially set out to just fix #535 but then realized that there was no frontend validation on the version creation page. So I tried to copy what the project creation page was doing and added in some frontend validation.


wafflecoffee (approved):
Since the Svelte rewrite is a bit on hold, let's merge this!
(after conflicts are resolved)

kevoh1516:
Yayyy. Will fix the conflicts this weekend.

PR #581, opened 2022-07-26
Fix the bunny fonts link from #580

frikinjay:


Akarys42:
There is no reason to rely on an external service when you have your own CDN. You shouldn't be adding unnecessary point of failures when you can avoid them, beside the loading time difference of having to connect to another, potentially slower, service.

Geometrically (approved):

frikinjay:
@Akarys42 I might create a separate pr for that.

wafflecoffee:
I've uploaded it to the CDN https://cdn.modrinth.com/fonts/inter/inter.css

PR #609, opened 2022-08-16
fixes Inconsistent donation icons and colours #608

Arbee4ever:
replaces all donation png's with svg's from https://simpleicons.org
removes svg files with "-svg" at the end

Note: The Ko-fi icon now has no circle around it, which is the only version simpleicons provides, in my opinion that is not a Problem as I think it looks better


wafflecoffee (approved):

Brawaru:
Maybe unknown donation platform should use [Coin icon coin](https://icon-sets.iconify.design/iconoir/coin/) or [Giving hands and heart in-between icon giving hands](https://icon-sets.iconify.design/iconoir/donate/)?

Arbee4ever:
> Maybe unknown donation platform should use [Coin icon coin](https://icon-sets.iconify.design/iconoir/coin/) or [Giving hands and heart in-between icon giving hands](https://icon-sets.iconify.design/iconoir/donate/)?

I'd probably use the giving hands.

wafflecoffee (approved):

PR #610, opened 2022-08-16
Add "other launchers" to warning cards

ubionexd:
Add "other launchers" because there are many other supported launchers referenced in the documentation, not just MultiMC, PolyMC, and ATLauncher.


DioEgizio:
https://docs.modrinth.com/docs/modpacks/playing_modpacks/

What are those "many other launchers"?

ubionexd:
Wait what-
I had these changes for like a month now but I made a pr now.
A month ago that page had like XMinecraft Launcher and other launchers like that.

Looks like those has been removed. Oops

PR #611, opened 2022-08-16
added support for the kbd tag in markdown editor

jadelily18:
* added kbd tag to whitelist in `plugins/xss.js`
* added styles for the kbd selector in `assets/styles/global.scss`

### DIFF
--- a/assets/styles/global.scss
+++ b/assets/styles/global.scss
@@ -51,6 +51,8 @@ html {
   --color-code-bg: var(--color-bg);
   --color-code-text: var(--color-text-dark);
 
+  --color-kbd-shadow: rgba(0, 0, 0, 0.25);
+
   --color-ad: #d6e6f9;
 
   --shadow-dropdown: 3px 3px 14px hsla(0, 0%, 0%, 0.15);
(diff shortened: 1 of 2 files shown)


wafflecoffee (approved):

Andre601:
To give some input, would it be a difficult task to implement characters for specific kbd cases?

Like when someone has `<kbd>cmd</kbd>` that it would display as <kbd>⌘ Cmd</kbd>?
Would IMO make the kbd a little bit better outside of simply adding minor styling to it...

Tho that's just an idea and not really a needed thing.

wafflecoffee:
Well, if someone wanted to have one that says ⌘, they would use ⌘. The reason this PR was opened is because we don't support it at all at the moment.

Andre601:
When I understand this PR right is this only adding the `<kbd>` HTML syntax to be usable, not any markdown-like formatting, right?

isXander (approved):
Yes please!

Minenash:
I didn't even know this tag existed. I've been using inline code blocks, this could be great!

PR #620, opened 2022-08-21
Fix search categories

Geometrically:


pages/search.vue
```diff
@@ -555,16 +557,23 @@ export default {
...
+          if (this.orFacets.includes('categories:bukkit'))
+            this.orFacets.push('categories:bukkit')
         } else if (elementName === 'categories:spigot') {
-          this.orFacets.push('categories:bukkit')
+          if (this.orFacets.includes('categories:bukkit'))
+            this.orFacets.push('categories:bukkit')
         } else if (elementName === 'categories:waterfall') {
-          this.orFacets.push('categories:bungeecord')
+          if (this.orFacets.includes('categories:bungeecord'))
+            this.orFacets.push('categories:bungeecord')
```
wafflecoffee, line 576: Isn't the logic on most of these if statements reversed? i.e., this line 576 is only adding the bungeecord orFacet if it already includes the bungeecord orFacet. It should be doing this if it does *not* already include that orFacet.

PR #622, opened 2022-08-22
Fix loaders not showing

Geometrically:


Andre601:
Can confirm that the issue is fixed by this change.

PR #627, opened 2022-08-25
Minor fixes for "See all" link design

Brawaru:
- Now styled as a proper link.
- Change :focus to :focus-visible to avoid link getting stuck in that 'hovered' state because focus does not actually change (it deffo should though, something to fix in the future - should be one fragment change).
- Removed all padding and transition.

<table>
  <tr>
    <th>State</th>
    <th>Light mode</th>
    <th>Dark mode</th>
  </tr>
  <tr>
    <td align="center">Normal</td>
    <td align="center">Preview of normal link state in light theme.</td>
    <td align="center">Preview of normal link state in dark theme.</td>
  </tr>
  <tr>
    <td align="center">Hovered</td>
    <td align="center">Preview of hovered link state in light theme.</td>
    <td align="center">Preview of hovered link state in dark theme.</td>
  </tr>
    <td align="center">Active</td>
    <td align="center">Preview of active link state in light theme.</td>
    <td align="center">Preview of active link state in dark theme.</td>    
  <tr>
  </tr>
</table>

### DIFF
--- a/pages/_type/_id.vue
+++ b/pages/_type/_id.vue
@@ -1047,15 +1047,18 @@ export default {
     display: inline-flex;
     align-items: center;
     gap: 3px;
-    padding: 5px 3px 5px 7px; // <- 3px & -> 7px to compensate for chevron
+
     border-radius: 5px;
-    transition: 0.05s all ease-in-out;
+    color: var(--color-link);
   }
 
   .all-link:hover,
-  .all-link:focus {
+  .all-link:focus-visible {
+    color: var(--color-link-hover);
+  }
+
+  .all-link:active {
     color: var(--color-link-active);
-    background: var(--color-card-link-bg);
   }
 }
 


Geometrically (changes requested):
I don't think should be the link color. No other button is like that on the site

Brawaru:
Well, it supposed to be a link. Should I change it back to normal colour (black/white depending on the theme) for normal state or for all states?

wafflecoffee:
> No other button is like that on the site

Maybe instead we make the other buttons on the site look like that. I think what's here looks a lot better and also makes a lot more sense in terms of accessibility. We could adopt something along the lines of what the rewrite has:
Peek 2022-08-25 11-12

jadelily18:
i like the idea of making it look more the the rewrite. It just looks really clean and easier to read compared to production.

Geometrically:
> > No other button is like that on the site
> 
> Maybe instead we make the other buttons on the site look like that. I think what's here looks a lot better and also makes a lot more sense in terms of accessibility. We could adopt something along the lines of what the rewrite has: Peek 2022-08-25 11-12 [ Peek 2022-08-25 11-12 ](https://user-images.githubusercontent.com/60205699/186703299-9ac58530-3308-4171-8d44-5b97a5f33ae9.gif) [ ](https://user-images.githubusercontent.com/60205699/186703299-9ac58530-3308-4171-8d44-5b97a5f33ae9.gif)

I don't think this is true. If it's styled as a button, it should look like a button. There shouldn't be a button-link hybrid. Even on the rewrite, buttons are never styled with a blue link color.

It should either be a link, which is blue with an underline, or a button with the background, hover, and normal text

Geometrically (approved):
thanks for the PR and putting up with all my bikeshed :+1:

PR #629, opened 2022-08-26
Try CarbonAds

### LINKED ISSUE #1882: It would be nice if the ad loading didn't push down content on the page

Right now, when the ad loads, it causes content under it to jump downwards.

https://user-images.githubusercontent.com/2068021/169670635-3a1f044a-2972-4e99-86d7-97f8bf62c12c.mp4

It would be nice if the ad did not cause content below it on the page to move. Maybe space could be reserved (with idk, margins or something) so that inserting the ad element does not affect the height of the page, or maybe the ad HTML could always exist and the text is just inserted into it.

_No response_

### Additional context

EthicalAds publisher policy does say "The ad should not disrupt the natural reading flow of the page."; it's referring to ad placement here, but this is kinda related? 😛

### LINKED ISSUE #2012: ad text does not update colour when changing theme
**Describe the bug**
Ad text does not change colour when the theme is changed, rendering it nearly unreadable.

**To Reproduce**
1. Go to a location with an ad, such as the [hydrogen mod page](https://modrinth.com/mod/hydrogen).
2. Swap theme.
3. See that the ad's text colour has not changed, making it difficult to read.

**Expected behavior**
The ad's text colour should have changed to match the current theme.

**Additional context**
This is a very minor issue, but I figured I'd just report it so it's known.

Geometrically:
As a step towards implementing author payouts, we're experimenting with a couple different ad providers to see which one works the best for us. We've been using EthicalAds for a long while now, but we'd like to now try out CarbonAds.

The ad banner is a bit larger than before, but it won't impact usability and it actually styles a lot better than Ethical does. We'll see how this works out for now and report back later with results!

Here's some screenshots:
image
image


jadelily18:
is that the same screenshot twice or am i missing something? i think it looks fairly good, just a bit plain for how large it is compared to other components

MMK21Hub:
That looks good: I like the blue background.

One thing that we'll need to be careful of is people viewing the site in (e.g.) landscape on a mobile. Will we be able to "squish" the ad so that the user can see a bit of content before they scroll down?

It looks like the entire as surface is clickable. Are the focus styles working properly?

Geometrically:
> That looks good: I especially like the blue background.
> 
> One thing that we'll need to be careful of is people viewing the site in (e.g.) landscape on a mobile. Will we be able to "squish" the ad so that the user can see a bit of content before they scroll down?
> 
> ~Finally, is it planned to do something about the jump in position of the content once the ad loads in? It was a slight annoyance with EthicalAds.~ _Already been reported: #488_

Yep! On the deploy the mobile version is the same height, with the text being smaller.

Geometrically:
> is that the same screenshot twice or am i missing something? i think it looks fairly good, just a bit plain for how large it is compared to other components

Thanks! I accidentally pasted the same screenshot twice. It's fixed now.

Akarys42:
This pull request breaks GDPR compliance. [Carbon is known to use trackers](https://www.ethicalads.io/images/posts/2021-third-party-resources.png), therefore you must get consent from users in the EEA before loading Carbon ads. 

Note that I'm not a lawyer and this isn't a legal advice.

FnigePython:
As per conversations in the discord server (which you should totally join :p) People seem to like it better if its above the main content bar (whatever its called)

Some pics of my hacky job

image

image

DioEgizio:
Minor detail but that should be round like everything else

FnigePython:
image
Further shenanigannery has lead to this

Also slight fix youll need to do (For anyone else wondering, carbon's rules require the image to be 130x100 pixels)
image

ix0rai:
> As per conversations in the discord server (which you should totally join :p) People seem to like it better if its above the main content bar (whatever its called)
this is inconvenient for the average user imo
it's best to have the search bar as close as possible to the top of the page, to let users click it and start typing faster

FnigePython:
So in summary for the discord thread stuffs so far:

Putting the ad banner above the search bar seems to be the most liked idea
- There is the competing idea of above the sidebar instead but that was mostly pushed literally by one single person lol
- Under the sidebar does not work because you would have to put only 2/3 categories up and the rest have to be behind a more button (And dont get me started on within project pages)

Removing the padding within the banner, meaning that the photo's height matches the ad banner's height seems to also be liked. This would lessen the ad's footprint by a total of ~14 pixels vertically

The ad picture is 117x90 and not 130x100 which should be fixed

And Mantaylia is starting to like web development a little too much (This is a cry for help)

Vetpetmon:
Some issues I've noticed on my end is that because CarbonAds uses trackers & cookies, adblockers are quick to block it. However, what's interesting is that it's not my opted-in adblocker here that's blocking the advertisements.

image

Opera GX applies a "blanket ban" on ads almost by default, which is great for security, but this also doesn't work out for supporting developers on Modrinth. As GX is becoming closer to being a standard browser, this could pose an issue since the adblocker isn't exactly opt-in. As CarbonAds is owned by BuySellAds, it's already (in)famous in many advert-blocklists, as it doesn't follow GDPR world-wide (It places trackers in all advertisements), only COPPA.

Additionally, what I think may be an issue is the old names I'm seeing in Inspect Element:
image

EthicalAds name fragments are still present, which may alter results. Consider trying to not name elements as advertisements.

In some cases, I'd suggest a failsafe of sorts: If CarbonAds can't operate, then we should try first-party monthly hosting deals for advertisements while avoiding certain key words adblockers look for when blocking ads. Independent investors/customers can pay per month to display ads that a whole audience will be able to see, while also avoiding the usage of trackers besides one that reports to the server how many times it's been loaded. A lot of adblockers will find this okay on their ends and not put it on a blocklist, and I'm in the process of testing such a method out myself.

This can even make advertisement more accessible for modders and artists on the platform, but keep in mind that this is a suggestion for a fail-safe when CarbonAds fails, and that can happen even without an adblocker (It's fully possible for their hosting servers to be down, lagging, hijacked, etc.).

_I applaud the team for sandboxing the ads by not using iframes._

Geometrically:
> This pull request breaks GDPR compliance. [Carbon is known to use trackers](https://www.ethicalads.io/images/posts/2021-third-party-resources.png), therefore you must get consent from users in the EEA before loading Carbon ads.
> 
> Note that I'm not a lawyer and this isn't a legal advice.

Just to reiterate from the discord: A Carbon rep has confirmed via email that Carbon is 100% GDPR-compliant. https://content.buysellads.com/publishers/what-data-does-buysellads-collect.

If you believe that CarbonAds is not GDPR compliant, feel free to contact your local authority to get action taken.

ZeoNight:
Oh no im so sorry i have a adblocker on-
image

Geometrically:
@ZeoNight How did that happen? Did you do anything specific?

ryanccn:
Something worth looking at

https://github.com/ryanccn/modrinth-ad-experiment

PR #635, opened 2022-08-28
Make the side filter menu always visible in desktop mode.

MrDrummer:
Makes the sidebar menu `position: fixed` when in desktop view.

Idea is that it's always visible when scrolling.

chrome_zMDJFHizBe

### DIFF
--- a/pages/search.vue
+++ b/pages/search.vue
@@ -847,6 +847,10 @@ export default {
 }
 
 @media screen and (min-width: 700px) {
+  .filters-card {
+    position: fixed;
+  }
+  
   .search-controls {
     flex-wrap: nowrap;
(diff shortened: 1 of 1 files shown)


Arbee4ever:
That would make it somewhat unusable on some screen resolutions, where it#s cut off at the bottom. Like on my 2nd monitor which is 1600x1200
image

MrDrummer:
What if the sidebar menu scrolls instead of the overflow being hidden, would that be considered user friendly?

Geometrically:
I'm going to close this PR as it seems to have worse UX and wouldn't work for the majority of MR's users. If you have a better idea/a good fix, feel free to reopen this. Thanks for the contribution though!

PR #661, opened 2022-09-12
More reliably detect ad load faults

Brawaru:
Detecting that ad has failed to load with timeout is not accurate and won't work on slow connections. This commit fixes that by changing detection to use 'onerror' event on <script> tag instead, which will be triggered if ad script fails to load due to blocking.


Geometrically (approved):
thanks for the PR!

PR #662, opened 2022-09-12
Move followed projects page to search

### LINKED ISSUE #464: Filter mod search page by followed mods

_No response_

I'd like to be able to filter mods by those I'm following on the mod search page:  
https://modrinth.com/mods

Alternatively, I'd like to have (the majority of) the same filter options on the "Followed Projects" page, but I feel this would be redundant.  
https://modrinth.com/settings/follows

### Additional context

I'd like this feature mainly to be able to sort the projects I'm following by "Recently updated". I know that I can already receive notifications when said projects are updated, but I use the same mods on multiple computers - rather than cloning my modded Minecraft instance every time I want to sync updated mods between computers, it would be nice to be able to simply view which mods have been updated since the "last modified" date of the mods folder, then reinstall the latest version of said mods.

(I'm sure there are better solutions for syncing instances between computers, but that's neither here nor there ;) )

It would also be nice to sort followed projects by client / server side, as well as by category.

triphora:
Resolves #464


Geometrically (changes requested):
TBH, I'm not a fan of the feature and I think the current followed projects page is better for following/unfollowing projects. 

The use case is really limited, and for most people it'll add clutter to the search bar. It also breaks sharing your searches by URL since it now depends on login state. For those reasons- I think this shouldn't be added to the site. (sorry if this came off as harsh)

wafflecoffee:
When I go to the search page, all I see are 20 mods which I'm already following and already aware of, so being able to exclude projects you're already following would be great for discovering new mods. Being able to filter through the projects you're already following is also a requested feature (see #464), and I think this is the best way to do it.

I don't think that sharing URLs is really an issue. If few people use it anyway like you say, they won't be sharing it that often with that either. I don't think that cluttering the search sidebar is an issue at all either, it barely extends past the edge of my screen.

Geometrically:
> When I go to the search page, all I see are 20 mods which I'm already following and already aware of, so being able to exclude projects you're already following would be great for discovering new mods. Being able to filter through the projects you're already following is also a requested feature (see #464), and I think this is the best way to do it.
> 
> I don't think that sharing URLs is really an issue. If few people use it anyway like you say, they won't be sharing it that often with that either. I don't think that cluttering the search sidebar is an issue at all either, it barely extends past the edge of my screen.

I think removing the followed page creates a lot worse UX. Having to go 4 different search pages to see the projects you're following and remove a follow would be really annoying.

With sharing URL's, even if it only breaks rarely, it still breaks. Search queries should be reproducible. 

I think this PR is more of an XY problem. The root issue here is discoverability of mods. Since this is the case, I think we shouldn't do a temporary fix for this, and address it to it's core, which would be something like creating a "Discover" page (similar to GitHub's) which would recommend new projects to people based on their choices (which we would do properly, protecting privacy and processing data correctly).

triphora:
Closing this because some alternative ideas for the followed projects system were brought up today in the Discord

PR #647, opened 2022-09-05
Rewrite Parity

Geometrically:
This PR aims to "take" features from the rewrite which were better than the Nuxt version. Since the rewrite is on hold, this is a good way to bring improvements from the rewrite to the Nuxt version.

Stuff that still needs to be done:
- [x] Cache tags at build time
- [x] Match User Page (including editing)
- [x] Add dashes to Changelog Page
- [x] Match Modal/Popup Styling
- [x] Correct meta for search pages
- [x] Navbar line animation
- [x] Allow admins to edit project data
- [x] New notifications page
- [x] Better ProjectCard view on mobile
- [x] Modal Creation Page (Done, Needs Feedback)
- [x] Better 404 Page
- [x] Rewrite Avatar Component (handle pixelated + replace placeholder)

Future work:
- Individual Version Page + Editing + Creation
- Game version search dialogue
- Gallery/project colors (?)
- Gallery Item Creation

PR #670, opened 2022-09-18
Add internationalization support

Brawaru:
> **Warning**
> This pull request message is still being worked on and is incomplete. Come back later.

This PR overhauls almost all parts of the site to add internationalisation support, thus making site fully localisable through existing localisation managing platforms, and more accessible to different speakers in most pasts of the world [^1] [^2].

# For maintainers

This PR touches almost all parts of the site. Big effort went towards making the migration seamless, while making as many parts localisable. Unfortunately, in certain cases compromises were made and changes to code, styles or layouts were affected, an in rare cases strings were changed too.

Introduced with this PR are multiple new APIs and components:

### Intl plugin

The plugin is providing basic APIs, those are storing the massages, current locale and IntlShape, as well as additional formatters. It however does not have any means of changing locales apart from re-initialising it with the new locale messages.

New APIs to Vue include:

- `$i18n` getter, which returns active `IntlController` instance. The controller is what manages all the locales and initialises the [IntlShape](https://formatjs.io/docs/intl#intlshape).
  - `intl` is a getter that returns the current IntlShape.
  - `locale` is a getter alias to `$i18n.intl.locale`, containing the current locale.
  - `intlLocale` is a getter that returns the `Intl.Locale` object for the current locale.
  - `formats` is a collection of formatter methods, these include most of the standard methods from `IntlShape` except for those that format to parts, as well as a few custom ones, namely:
    - `compactNumber`, which receives a number and options on how to format it, then formats the provided number in compact notation and returns array of the formatted string and rounded number that you can use to pluralise the strings. The explanation is provided below.
    - `timeDifference`, which receives the ending date and the start date (optionally and if not provided, considers it to be the current day) under the hood it calculates the amount of time between the two dates, and then automatically chooses the most appropriate unit to dispay (e.g. for 16 hours difference it would make sense to say 16 hours, not 57,600 seconds), it then calls the `formatRelativeTime` function and returns the formatted value.
    - `customMessage`, which receives the message, values and the formatter options. It then formats the message and returns the result. Essentially it's more easily accessible [Intl MessageFormat API](https://formatjs.io/docs/intl-messageformat).
- `$fmt` is an alias getter to `$i18n.formats`.
- `$t` is a function that receives message ID and values to format it which, it is essentially the alias to `$i18n.intl.formatMessage({ id }, values)`.

<details>
<summary>Why <code>compactNumber</code> API matters?</summary>

_[From the introductory commit `1fde295`](https://github.com/Brawaru/knossos/commit/1fde2950bf9e6edf4bb33081cd2f075376586c94)_

Unfortunately, Intl.NumberFormat API has a serious weakness when it comes to using the numbers in compact notation in translations: you cannot choose correct plural form based on the result it provides. That's because the result is a completely formatted string, which you cannot reliably parse for all the locales.

Many consider that passing the original number for plural selection will work, but that's not the case, because languages actually have declensions for the words coming before or after the number in compact notation. Not to mention, the very number for compact unit can decline based on the number in compact notation, which is different from the original number.

For example, here's how English cats translate to Ukrainian:

- 0.5 cats ⇒ 0,5 кота.
- 1 cat => 1 кіт.
- 2 cats => 2 коти.
- 5 cats => 5 котів.
- 1 thousand cats => 1 тисяча котів.
- 2 thousand cats => 2 тисячі котів.
- 5 thousand cats => 5 тисяч котів.

Pay close attention to how Ukrainian changes when talking about thousands. Nothing magical happening here: it's just that instead of cats, it starts to decline word for 'thousand' instead, and word for 'cats' goes into genitive plural case.

Now, what would happen if we were to use compact notation?

- 0.5 cats ⇒ 0,5 кота.
- 1 cat => 1 кіт.
- 2 cats => 2 коти.
- 5 cats => 5 котів.
- 1K cats => 1 тис. котів.
- 2K cats => 2 тис. котів.
- 5K cats => 5 тис. котів.

English just changed to use 'K' after the number, yet Ukrainian kept the word for 'thousand' there, just shortened it ever so slightly. Yes, the word is still here, so keep that in mind.

It's important that I gave you example for small numbers, including decimals, because they demonstrate all four plural forms Ukrainian has.

Let's write that cat format using ICU Message Format:

```icu
{cats, count, one {# cat} other {# cats}}
```

Yep. It's that simple in English because it uses just two plural forms, 1 (and -1, but that's controversial) for 'one', and anything else goes to 'other'.

Now let's translate it to Ukrainian:

```
{cats, count, one {# кіт} few {# коти} many {# котів} other {# кота}}
```

Nothing suspicious yet.

But then let's use this translation with the following numbers of cats:

- 1,500 ⇒ 1,5 тис. котів.
- 1,501 ⇒ 1,5 тис. кіт.
- 1,502 ⇒ 1,5 тис. коти.
- 1,505 ⇒ 1,5 тис. котів.

If you've read the above carefully, you already spotted the mistake. This is the weakness of JavaScript Intl API. If you correctly use ICU library for Java, it will, in fact, correctly apply the plural, because when you convert number to compact notation in ICU it does not give you the formatted string, it gives you a special object that you can then use as context value to format your message. Intl.NumberFormat in JavaScript, unfortunately, returns plain string, and even formatting to parts can't help you.

So how do we go to fix this problem? The writers of the original Intl.NumberFormat proposal already work on finding a way to mitigate this issue, but it will take time, it already took enormous amount of time while sites continue fall for this pitfall.

The solution is not easy: you need to use original CLDR data to correctly calculate number on which you should determine the plural form. Glad for us, there's a polyfill for Intl.NumberFormat with its internals somewhat exposed. We can use — abuse, really — these internals to perform what Intl does behind the scenes: calculate exponents and actual number value.

Unfortunately, this comes with a significant drawback: you must duplicate context values twice: once for formatted number (in compact notation), and then for number that you actually use to select the plural form. This can be somewhat confusing to translators, but it's what we got to do to mitigate the obvious mistake.

I created a simple formatCompactNumber function within IntlShape, which accepts the value to format and almost all — except for notation, obviously — number formatting options; it then returns a tuple that contains the formatted number and number for pluralisation.

Because it's a tuple, we'd want to give it a meaningful value, for what I created another helper function — deunionize. It accepts a tuple, or any array, really, and keys as parameters. It then creates hashmap (an object without any prototype) for every key assigns value from array, in order of passed keys.

Thus, we can do this when translating strings that use counters:

In translation file,

```json
{
  "downloads": "{downloads, plural, one {{downloadsCounter} download} other {{downloadsCounter} downloads}}"
}
```

In code,

```js
$t(
  'downloads',
  $deunionize(
    $fmt.formatCompactNumber(downloads),
    'downloadsCounter',
    'downloads'
  )
)
```

deunionize function will create us the following hashmap (for 1,501):

```js
;({
  downloadsCounter: '1,5K',
  downloads: 1_500,
})
```

</details>

It also provides the new `IntlFormatted` component.

<details>
<summary>IntlFormatted explanation</summary>

_[From the introductory commit `7be6377`](https://github.com/Brawaru/knossos/commit/7be63779814de9635f8c6b741902c17b32a24d89)_

IntlFormatted is a rewritten i18n-formatted component that uses slots instead of the magical directive, providing more verbose, but cleaner templating, without weird magic. It also gives ability to create nested wrappers because children are now passed as slot bindings rather than inserted directly into wrapper element.

Here's an example usage of the new component. It demonstrates most of the `IntlFormatted` features, but you probably won't use it like that (e.g. you'll merge `target` and `~targetName` slots, etc).

```html
<IntlFormatted
  message="<i>Hello, <target>{targetName}</target>!</i>"
  :tags="['i']"
  :values="{ $name: 'Brawaru' }"
>
  <template #target="{ children }">
    <strong><Fragment :of="children" /></strong>
  </template>
  <template #~targetName="{ values: { $name } }"> {{ $name }} </template>
</IntlFormatted>
```

There are a few things to unpack in this example:

1. We provide a custom message to the component, in that message we use both values (`targetName`) and wrappers (`i`, `target`).

2. We then define standard tags that will be provided automatically, namely `i`, an element for italic text.

3. We then provide values to format message with. Because we don't want `$name` value to be used in texts, as it is a computed property, we prefix its name with `$`, which would make it invalid key name within ICU Message Format syntax.

4. Then we define template for `target` slot. Slot names can be anything, they will be used as a name for the wrapper. For wrapper slots we can expect the properties `children` (formatted contents inside the wrapper) and `values` (the initially provided values to format message with).

   > **Note**
   >
   > It is recommended to use values in the formatted message when possible instead of using the one from object. They are provided for the convenience of accessing computed values (such as compact numbers).

5. Finally, we also define template for `~targetName` slot. Notice `~` character in the slot name. All tilde slots are provided as a value for formatting the message, not as a wrapper. Therefore it will not have any `children` in slot properties, only `values`.

The above component will render to this HTML:

```html
<i>Hello, <strong>Brawaru</strong>!</i>
```

As you can see, the new component is quite powerful, verbose, but easier to understand of what's going on. Previous component is now deprecated and will be removed in the following commits, as soon as all of its usage will be replaced with the new component.

</details>

### Nuxt module

What allows to change the locales is the Nuxt module, which extends the plugin to provide functions to dynamically load and change locales, as well as automatic mode, where locale can be driven by the browser own preference, which is made the default. The module loads all the locale files based on the `nuxt.config.js` configuration, it then generates import map where locale files are split in multiple chunks to avoid sending users too much data at the same time. Additionally, all the Vue helpers are also injected in Nuxt context, app and store to be accessible from there.

Extends include:

- `$i18n`:
  - `availableLocales` is an array of locale descriptors, which are objects that include `code` (locale code) and `data` (optionally provided meta data about the language available without loading the language; useful for storing things like custom locale name (which is exactly how we use it)).
  - `changeLocale` is a method that allows to change the current locale, it accepts locale code as a first argument and optionally boolean value that dictates whether the locale preference must be stored (it's `true` by default). Locale code can be a special string `'auto'`, which removes the preference, detects the navigator locale and changes locale to it without saving, and then also turn on automatic mode.
  - `automatic` is a getter that returns boolean value whether the automatic mode is currently active. If it's `true`, then locale is changed according to browser settings, in real time. To turn off this mode you cannot set `automatic` to `false` (there is no `automatic` setter), instead you should call `changeLocale` with any existing locale code.
  - `data` is a getter that returns imported data associated with the locale. See Nuxt configuration for explanation how imported data works.
- `$i18nHead` is a function that will return an initial `vue-meta` object containing important meta tags that allow search engines to better understand how to index the site in different languages, as well as changes page locale for accessibility reasons.

#### Module configuration

As said above, the module is configured through `nuxt.config.js`, via either `i18n` root property or module settings. If both are provided, then module options will override the options from root property.

The following options are available:

- `defaultLocale`, a string with default locale code, it will be used if module fails to detect the navigator language, or no locale is found to fit the navigator preferred languages.
- `localesDir`, a string with path to the directory that contains all the locale files. Locale source files or imported data cannot be provided outside this directory.
- `baseURL`, an optional string containing the site base URL (domain), without it the meta tags for search engines cannot be generated.
- `locales`, an array of locale descriptors, object containing information on how to load locale, its custom metadata and data imports. Each descriptor has the following properties:
  - `code`, a string with locale code (e.g. `de`, `fr-CA` or `en-x-custom`). It must be a canonical BCP47 tag, but it may be invalid (although that is not recommended). Use codes wisely, in many cases you don't have to provide the region, which would limit the locale usage to that region, or you need to provide locale script rather than region, e.g. `zh-Hans`. Please read [this article by W3C](https://www.w3.org/International/questions/qa-choosing-language-tags) to gather more understanding on how to choose locale codes.
  - `file`, a string with relative to `localesDir` path to the locale source file.
  - `additionalImports`, is an optional array of import paths that will be imported together with the locale source file. The result of the improt will not be accessible in runtime unless the imported file declares properties globally. This is useful to import data for polyfills since shipping all data for all locales at the same time can have huge costs on the bundle size.
  - `importedMap`, an optional object where objects of either import path or array of import path and import key are mapped by any keys under which they can be found in runtime under the `$i18n.data`. If only import path is used, then the import key is considered to be the `default`. This option is useful if you want to import additional resources associated with the locale. All the resources must be located in the `localesDir`.
  - `data`, is an optional object containing all the metadata associated with the locale. That metadata will be available even before the locale has been loaded, which makes it great way to define properties that must be always available, like custom locale names, flags or anything else.

### Webpack loaders

JSON is a decent format to provide locale messages, but it becomes pain to write manually. Unfortunately, we do not have automatic extraction, so we have to write our locale files by hand. The decision has been made to choose TOML for source files. Additionally, some long strings were split to the separate resource files.

On top of all that, since we use Intl MessageFormat, it would make sense for us to pre-parse all the messages to avoid the cost of doing that in runtime (twice (!) for the SSR and FE). While parsed messages are obviously have their toll on the bundle size, they should compress relatively well. You are encouraged to read the article from FormatJS that provides [more information about parsing of the messages](https://formatjs.io/docs/guides/advanced-usage).

This is why two loaders were introduced to the configuration. All the loaders are located in `locales` directory. These two loaders are:

- `messageLoader`, which acts like `raw-loader`, except it parses the contents as a message and exports the parsed result.
- `translationsLoader`, which loads our source TOML files, flattens the keys, parses every message from them and returns the result.

### Locale files

As told above, we have source files in TOML, for our configuration we also have a mandatory meta file `meta.json` with each locale, it is never exported to the locale, however properties from it used in locale datas.

Additional resources are supported as well, and we have two of them already (`revoke-token.html` for the long message in revoke-token modal, and `starter.md`, which contains starter descriptions for the projects to guide users on how to set up their project).

Any additional files will be automatically picked up and added to `importedData`, under the key that matches the file name (including extension). In some cases you will need to tell webpack how to process the file: if you use it as a message, then process it as one via `messageLoader`, if it's just text that won't be formatted, then use built-in `raw-loader`.

</details>

# For translators

As a translator myself I tried to make experience for us translators as good as I possible can. That included adopting formatJS libraries working in the current Vue frontend instead of using nuxt-i18n with vue-i18n. That means we are getting the ICU MessageFormat that we are used to already, with additional features like HTML-like ‘wrappers’ (aka rich text formatting).

formatJS have [a good summary of ICU MessageFormat syntax](https://formatjs.io/docs/core-concepts/icu-syntax) available on their site. If you are not familiar with this syntax, I encourage you to read the summary. It's pretty simple syntax that gets the job done. At our hands are `{variables}` (aka arguments), `{plurals, plural, one {# plural} other {# plurals}}}` (where `#` will be replaced with the number, but see the note below), `{selectors, select, are_cool {are cool} other {are so-so}}`, dates and other formats.

> **Note**
>
> While we're using plurals, in cases of numbers in compact notation (e.g. 1.2K), the number used as plural argument is not a real number: it's rounded to match the formatted number. The actual number will be available as a string via a separate argument.

‘Wrappers’ as I got used to call them, or rich text formatting as formatJS call them, are special syntax elements that allow using tags on the text to format it in the result, these changes often are purely visual.

You don't need to know any HTML to use them, just know that for one tag there's start tag, which contains the tag name surrounded by ‘less than’ and ‘greater than’ characters (`<my-tag>`) and end tag, which almost the same, but has slash before the name (`</my-tag>`), everything in-between the two is considered contents of that element (`<my-tag>This is the content! You can use any {variable}s and even other <bold>tags</bold> here.</my-tag>`). Just like with ICU variables, you can disable tags by surrounding them in quotation marks.

Crowdin should help you to cut the noise by reducing the names to zero-indexed numbers. E.g. `<strong>Warning:</strong> you are now warned. <italic>Please be mindful of that</italic>.` will become `<0>Warning:</0> you are now warned. <1>Please be mindful of that</1>.`.

> **Note**
>
> You still can see how Crowdin will interpret short tags by hovering over them with your mouse. If you click on that tag, it will be added to translation input field.

To unfortunate news, Crowdin does not currently support taking comments from TOML files into context, therefore I was unable to document variables used in the strings. After files are uploaded, I will try to add screenshots and even edit context for some strings that use variables, but otherwise variable names should represent well what values they hold, but don't fear to ask in Crowdin comments for additional context.

## Merging

This pull request can only be merged after the [`rewrite-parity` branch is merged](https://github.com/modrinth/knossos/pull/647), since it is based on that branch. Squash merge is preferred merging strategy, but the original branch will not be deleted to preserve the history of changes and explanatory commit messages.

## Credits

Huge inspiration was taken from [`nuxt-i18n`](https://github.com/nuxt-community/i18n-module) module and [`vue-i18n`](https://github.com/kazupon/vue-i18n/) plugin APIs. Since this is the the first time I work with Nuxt and Vue, their libraries provided guidance of how changes can be implemented. Big thanks to [formatJS authors and contributors](https://formatjs.io/) for providing libraries that power this contribution.

**Support my work**. This contribution took time and energy, so if you appreciate my efforts, please consider [supporting me](https://github.com/Brawaru/Brawaru/blob/main/SUPPORT.md).

[^1]: We currently do not support Right-to-Left languages as we cannot gurantee proper support for these due to maintanence costs that we are not willing to take.
[^2]: Content on the site is still provided mostly in English language and team is working to establish rule banning sole usage of other languages in the project descriptions, requiring descriptions to be provided in English with additional languages possible to be added through usage of `<details>` tags or additional links. There's an [issue about adding translatable project descriptions](https://github.com/modrinth/labrinth/issues/359#issue-1251962600) in Labrinth (our backend) to improve this experience.


i18n/nuxt/en-US.json
```diff
@@ -0,0 +1,38 @@
+{
```
Brawaru, line 1: Note: these files supposedly should go to https://github.com/modrinth/translations later.

Brawaru:
Language persistence added ✨ 

I think it's mostly ready API-wise, so a quick review of this monstrosity would be appreciated. In the following days probably going to refactor it a little bit (move utility methods around, check documentation, etc), rebase on rewrite parity branch and start replacing all the strings and preparing the initial translation file (it's *not* going to share translations with rewrite).

Brawaru:
Also regarding failing CI: why does it use Node.js 12 💀 Current LTS is 16.17.0

triphora:

PR #672, opened 2022-09-20
Adopt vue-prism-editor over CodeMirror

jadelily18:
This PR serves as an alternative to CodeMirror with a smaller bundle-size using [vue-prism-editor](https://github.com/koca/vue-prism-editor). Although that repository seems to be abandoned, it still serves as a good alternative, using [highlight.js](https://highlightjs.org/), which is already being used in knossos. It's also built on @Geometrically's rewrite-parity branch, so that it will be ready when that's merged.

This includes a MarkdownEditor component, only implemented in the project editing page for now.

#### Known bugs
* First of line numbers likes to show as 999 instead of 1, [only on Nuxt](https://codesandbox.io/s/nuxt-vue-prism-editor-62e06?file=/pages/index.vue) for some reason.


triphora:

PR #678, opened 2022-09-26
Bump @nuxtjs/color-mode to v2.1.1

Brawaru:
This does not affect the site in any way except that the color preference is no longer stored using the cookies, it's only localStorage now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "xss": "^1.0.8"
       },
       "devDependencies": {
-        "@nuxtjs/color-mode": "^1.1.1",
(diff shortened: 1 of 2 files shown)


wafflecoffee (approved):

Brawaru:
Just remembered something that I needed to check (whether it will work fine on my i18n branch[^1] if it's going to remain in `buildModules`). Everything seems aight, so this is ready for merge.

[^1]: Since this is where the change comes from: the current 1.x version does not work with dynamically generated head for `vue-meta`, and I really need it to properly localise the site.

PR #681, opened 2022-10-02
Fix 404 error when creating a resource pack

Brawaru:


components/ui/ModalCreation.vue
```diff
@@ -166,7 +166,7 @@ Questions? [Join the Modrinth Discord for support!](https://discord.gg/EUHuJHt)`
         })
 
         this.$refs.modal.hide()
-        await this.$router.replace(`/${projectType.display}/${this.slug}`)
+        await this.$router.replace(`/${projectType.actual}/${this.slug}`)
```
Brawaru, line 169: Just to note: there's no need for handler for `plugin` type (I tried to add it...): since created project won't have any versions it will immediately be considered of the parent type — `mod`; handling that would just create a plugin → mod redirect and unwanted blink due to that.

triphora:

PR #683, opened 2022-10-03
Fix special characters in gallery title/description

e-im:
Fixes modrinth/knossos#684

### DIFF
--- a/pages/_type/_id/gallery.vue
+++ b/pages/_type/_id/gallery.vue
@@ -345,8 +345,9 @@ export default {
               : null
           }&featured=${item.featured}`
 
-          if (item.title) url += `&title=${item.title}`
-          if (item.description) url += `&description=${item.description}`
+          if (item.title) url += `&title=${encodeURIComponent(item.title)}`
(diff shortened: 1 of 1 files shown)


e-im:
I had wondered the same thing, but saw it used in some other places as well (although, taking a look now, that was search— not sure how you're handling that in the backend/if its fed directly to meilisearch).

I'm happy to look into using PATCH body and pr that

wafflecoffee:
I imagine it was done like that in order to work around actix/actix-web#2695, which has been an absolute pain for us ever since we upgraded to actix 4.0. It might be possible to give the option to allow API users to either do standard JSON body or URL parameters, not entirely sure. It also might be better to wait until API v3 for that, development on which will likely start late this year or early next year.

The places to change would be [here for the POST route](https://github.com/modrinth/labrinth/blob/3e52f804/src/routes/projects.rs#L1174) and [here for the PATCH route](https://github.com/modrinth/labrinth/blob/3e52f804/src/routes/projects.rs#L1301).

PR #700, opened 2022-10-23
Persistent version filters

### LINKED ISSUE #603: Create URL for filters in a project

When you filter in a project on version or loader. Then select a version and then go back to the main version page, the filter is reset. Forcing you to set the filters again to your needs.

firefox_RyAbCysiHa

Creating a query parameter like `/mod/test/versions?filter=bukkit,spigot`. Or something similar.

This would also make it easier to share the correct version with users, for example if you have a project that is a plugin as well as a mod and you want to link it lets say on spigot, people usually would like to download the plugin, so it would be nice if the version tab already filters by CraftBukkit/Spigot

_No response_

### Additional context

_No response_

henkelmax:
This PR adds two query parameters to the *Versions* and *Changelog* sections.
Selecting and unselecting loader and Minecraft versions dynamically updates the query parameters.
Refreshing the page or sharing the link with filters active will automatically apply the filters.
Nonexistent loaders or Minecraft versions will automatically be removed on page load.

https://user-images.githubusercontent.com/13237524/197406455-91a404d9-5bc0-4774-8d25-f2da74c7432e.mp4

This closes #603

### DIFF
--- a/components/ui/VersionFilterControl.vue
+++ b/components/ui/VersionFilterControl.vue
@@ -48,6 +48,7 @@
       label="Include snapshots"
       description="Include snapshots"
       :border="false"
+      @input="updateQuery"
     />
     <button
       title="Clear filters"
@@ -94,11 +95,6 @@ export default {
       selectedGameVersions: [],
       selectedLoaders: [],
     }
-  },
-  watch: {
-    showSnapshots() {
-      this.updateQuery()
-    },
   },
   mounted() {
     this.selectedLoaders = this.getQueryAsArray(this.$route.query.loaders)
--- a/components/ui/VersionFilterControl.vue
+++ b/components/ui/VersionFilterControl.vue
@@ -95,11 +95,17 @@ export default {
       selectedLoaders: [],
     }
   },
+  watch: {
+    showSnapshots() {
+      this.updateQuery()
+    },
+  },
   mounted() {
     this.selectedLoaders = this.getQueryAsArray(this.$route.query.loaders)
     this.selectedGameVersions = this.getQueryAsArray(
       this.$route.query.gameVersions
     )
+    this.showSnapshots = this.$route.query.showSnapshots === 'true'
     this.updateVersionFilters()
   },
   methods: {
@@ -168,6 +174,7 @@ export default {
             ...this.$route.query,
             loaders: this.getAsQuery(this.selectedLoaders),
             gameVersions: this.getAsQuery(this.selectedGameVersions),
+            showSnapshots: this.showSnapshots ? true : undefined,
           },
         })
         .catch(() => {})
--- a/components/ui/VersionFilterControl.vue
+++ b/components/ui/VersionFilterControl.vue
@@ -97,11 +97,9 @@ export default {
     }
   },
   mounted() {
-    this.selectedLoaders = this.getQueryAsArray(this.$route.query.loaders)
-    this.selectedGameVersions = this.getQueryAsArray(
-      this.$route.query.gameVersions
-    )
-    this.showSnapshots = this.$route.query.showSnapshots === 'true'
+    this.selectedLoaders = this.getQueryAsArray(this.$route.query.l)
+    this.selectedGameVersions = this.getQueryAsArray(this.$route.query.g)
+    this.showSnapshots = this.$route.query.s === 'true'
     this.updateVersionFilters()
   },
   methods: {
@@ -166,9 +164,9 @@ export default {
         .replace({
           query: {
             ...this.$route.query,
-            loaders: this.getAsQuery(this.selectedLoaders),
-            gameVersions: this.getAsQuery(this.selectedGameVersions),
-            showSnapshots: this.showSnapshots ? true : undefined,
+            l: this.getAsQuery(this.selectedLoaders),
+            g: this.getAsQuery(this.selectedGameVersions),
+            s: this.showSnapshots ? true : undefined,
           },
         })
         .catch(() => {})
--- a/components/ui/VersionFilterControl.vue
+++ b/components/ui/VersionFilterControl.vue
@@ -95,7 +95,23 @@ export default {
       selectedLoaders: [],
     }
   },
+  mounted() {
+    this.selectedLoaders = this.getQueryAsArray(this.$route.query.loaders)
+    this.selectedGameVersions = this.getQueryAsArray(
+      this.$route.query.gameVersions
+    )
+    this.updateVersionFilters()
+  },
   methods: {
+    getQueryAsArray(query) {
+      if (Array.isArray(query)) {
+        return query
+      } else if (typeof query === 'string') {
+        return query.split(',')
+      } else {
+        return []
+      }
+    },
     getValidVersions() {
       if (!this.cachedValidVersions) {
         this.cachedValidVersions = this.$tag.gameVersions.filter((gameVer) =>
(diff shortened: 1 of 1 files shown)


triphora (approved):
Wow, I'm surprised that this doesn't conflict with the changes in the rewrite-parity branch!

henkelmax:
> Could you commit again? It's not letting me deploy to vercel

Sorry I won't be on my pc for the next couple of days.

Prospector (approved):

henkelmax:
> Could you commit again? It's not letting me deploy to vercel

@Geometrically Hope that fixes it

components/ui/VersionFilterControl.vue
```diff
@@ -95,7 +95,29 @@ export default {
       selectedLoaders: [],
     }
   },
+  watch: {
+    showSnapshots() {
+      this.updateQuery()
+    },
+  },
```
Geometrically, line 102: improper use of watch, this should use @change on the checkbox

components/ui/VersionFilterControl.vue
```diff
@@ -131,8 +154,37 @@ export default {
...
+          },
+        })
+        .catch(() => {})
+    },
+    getAsQuery(elements) {
+      if (elements.length === 0) {
+        return undefined
+      }
+      return elements.join(',')
+    },
```
Geometrically, line 187: This shouldn't be a separate function if it is only called once. The parent function should be `async` and await this promise instead of having an empty catch block

components/ui/VersionFilterControl.vue
```diff
@@ -131,8 +154,37 @@ export default {
...
+    validateFilters() {
+      this.selectedLoaders = this.selectedLoaders.filter((loader) =>
+        this.getValidLoaders().includes(loader)
+      )
+      this.selectedGameVersions = this.selectedGameVersions.filter((version) =>
+        this.getValidVersions().some(
+          (validVersion) => validVersion.version === version
+        )
+      )
+    },
```
Geometrically, line 169: also shouldn't be a seperate func if only called once

components/ui/VersionFilterControl.vue
```diff
@@ -95,7 +95,29 @@ export default {
...
+  },
   methods: {
+    getQueryAsArray(query) {
+      if (Array.isArray(query)) {
+        return query
+      } else if (typeof query === 'string') {
+        return query.split(',')
+      } else {
+        return []
+      }
```
Geometrically, line 119: is there a reason why both of these cases are supported? IMO for simplicity the version filters should always be in one format. I think just keeping it as a JS array is best (since this is what is done when it is serialized)

components/ui/VersionFilterControl.vue
```diff
@@ -131,8 +154,37 @@ export default {
...
+      )
+    },
+    updateQuery() {
+      this.$router
+        .replace({
+          query: {
+            ...this.$route.query,
+            loaders: this.getAsQuery(this.selectedLoaders),
+            gameVersions: this.getAsQuery(this.selectedGameVersions),
+            showSnapshots: this.showSnapshots ? true : undefined,
```
Geometrically, line 177: these query params should be shortened to just `l`, `g`/`gv`, and `s` (to make the url shorter and be like search

components/ui/VersionFilterControl.vue
Geometrically: ignore that this comment contains `getAsQuery`, accidentally selected too much

Geometrically, line 187: just for code quality I think this func is fine being inlined with a ternary in both places

components/ui/VersionFilterControl.vue
```diff
@@ -95,7 +95,29 @@ export default {
...
+    },
+  },
+  mounted() {
+    this.selectedLoaders = this.getQueryAsArray(this.$route.query.loaders)
+    this.selectedGameVersions = this.getQueryAsArray(
+      this.$route.query.gameVersions
+    )
+    this.showSnapshots = this.$route.query.showSnapshots === 'true'
+    this.updateVersionFilters()
+  },
```
Geometrically, line 110: this should be `async fetch()` so it works with ssr

components/ui/VersionFilterControl.vue
```diff
@@ -95,7 +95,29 @@ export default {
       selectedLoaders: [],
     }
   },
+  watch: {
```
Geometrically, line 98: (ignore)

Geometrically (changes requested):
I meant to hit request changes on my earlier review (updating this so PR status updates). See my comments above

components/ui/VersionFilterControl.vue
henkelmax: > This shouldn't be a separate function if it is only called once.

It is called by `updateVersionFilters` and from the include snapshots checkbox.

henkelmax: > The parent function should be async and await this promise instead of having an empty catch block

Since the `updateQuery` is now also called in the event of the checkbox instead of the watch, you can't await it there.

components/ui/VersionFilterControl.vue
henkelmax: The query string is only represented as an array if there are multiple query parameters with the same key:

**This returns an array**
`loaders=forge&loaders=fabric`

**This returns a string**
`loaders=forge,fabric`

So we have to handle both cases, since both of them are possible.

components/ui/VersionFilterControl.vue
Geometrically: > > The parent function should be async and await this promise instead of having an empty catch block
> 
> Since the `updateQuery` is now also called in the event of the checkbox instead of the watch, you can't await it there and creating a separate method for that doesn't seem to be reasonable.

This isn't how Vue does stuff, you don't have to await in an event for it to still be run async. You should make it async and await it- that's the codestyle this repo uses

components/ui/VersionFilterControl.vue
Geometrically: > The query string is only represented as an array if there are multiple query parameters with the same key:
> 
> **This returns an array** `loaders=forge&loaders=fabric`
> 
> **This returns a string** `loaders=forge,fabric`
> 
> So we have to handle both cases, since both of them are possible.

I'm confused how they are both possible. That doesn't make sense. The UI is what creates the URLs which are shared with people. Does the UI create both of these instances, if so, why? If it doesn't, then the code that handles what the UI doesn't "create" in the URL should be removed. We should only handle supported cases in URL params. Otherwise it introduces complexity which isn't needed. (see how search doesn't need this)

components/ui/VersionFilterControl.vue
```diff
@@ -131,8 +156,27 @@ export default {
...
+                ? undefined
+                : this.selectedLoaders.join(','),
+            g:
+              this.selectedGameVersions.length === 0
+                ? undefined
+                : this.selectedGameVersions.join(','),
+            s: this.showSnapshots ? true : undefined,
+          },
+        })
+        .catch(() => {})
```
Geometrically, line 178: should be async/await instead of a one line catch

Geometrically (changes requested):

components/ui/VersionFilterControl.vue
henkelmax: The array case is only if the user creates the URL manually with the same query parameter multiple times. But that case still needs to be handled, since it would lead to an error otherwise.

Geometrically: >  only if the user creates the URL manually 

Users should not be doing this and we should not support this case

Geometrically (changes requested):

Geometrically (approved):
Thanks for the PR! Lgtm

Mysterious-Dev (approved):

PR #719, opened 2022-11-11
Add more footer links

triphora:
image

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -367,9 +367,20 @@
       <div class="links links-2" role="region" aria-label="Resources">
         <h4 aria-hidden="true">Resources</h4>
         <a target="_blank" href="https://blog.modrinth.com">Blog</a>
-        <a target="_blank" href="https://discord.gg/EUHuJHt">Discord</a>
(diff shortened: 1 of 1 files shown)


layouts/default.vue
```diff
@@ -367,9 +367,20 @@
...
+        <a target="_blank" href="https://status.modrinth.com">Status</a>
+        <a target="_blank" href="https://github.com/modrinth">GitHub</a>
+      </div>
+      <div class="links links-3" role="region" aria-label="Interact">
+        <h4 aria-hidden="true">Interact</h4>
+        <a target="_blank" href="https://discord.gg/EUHuJHt">Discord</a>
+        <a target="_blank" rel="me" href="https://floss.social/@modrinth">
+          Mastodon
+        </a>
+        <a target="_blank" href="https://twitter.com/modrinth">Twitter</a>
```
Geometrically, line 380: These should be switched, Twitter is more relevant to the majority of people (since a lot more people have twitter than mastodon)

Geometrically, line 380: These should be switched b/c more people use Twitter than Mastodon (so Twitter is a more relevant link)

Geometrically (changes requested):

PR #727, opened 2022-11-13
Bump Vue to 2.7

Brawaru:
List of changes in Vue 2.7:
https://blog.vuejs.org/posts/vue-2-7-naruto.html

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3388,6 +3388,39 @@
         "@babel/core": "^7.0.0-0"
       }
     },
+    "node_modules/@vue/compiler-sfc": {
+      "version": "2.7.14",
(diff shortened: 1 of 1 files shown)


Brawaru:
Switch to `defineComponents` and stuff will be made with TypeScript, for now it's just a bump without any breaking changes (?). You definitely should test that though. The reason to upgrade now is that it will help me a bit when it comes to work on i18n stuff.

Btw, if you run `npm upgrade` as is, there are so many other changes (mainly babel and stuff):
```
added 10 packages, removed 56 packages, changed 214 packages, and audited 1473 packages in 2m
```

triphora (approved):
Tested locally and it seems to have no issues/regressions/etc

PR #732, opened 2022-11-13
Add fun little comment

Andre601:
Asked Emma on Discord if that can be made.
We should aknowledge the ~~nerds~~ people checking the code ;)


Andre601:
Checking the now made previews, looks like comments get removed... rip

triphora:
Makes sense, but that sucks.

Andre601:
> Makes sense, but that sucks.

At least we tried.

PR #735, opened 2022-11-15
New ads (same size + layout)

Geometrically:
How it looks:
image
image
image
image


HyperSoop:
this makes me wanna buy a server from exaroton lul
doesn't it make it more intrusive that the image adds to the ad's height on mobile?

Brawaru:
What's Vercel link to test this? I have a suspicion that leaving EA's data attributes allows ad blockers to do their work.

DioEgizio:
https://knossos-git-new-ads-modrinth.vercel.app seems to work

Brawaru:
Yeah, disregard: seems that no popular blocking lists block the EA attributes.

HyperSoop:
So, will it be adjusted to be in line with how Ethical looks on mobile? The image should be moved to be in the same line as the text instead of creating another one on top.

semi-related: you can't see a single mod card fully! Mod cards overall look weird. They need adjustments to be more compact, like reductions to the amount of text in favour of simple icons, decreasing the size of the icon, etc. you could throw up something like Mod Menu's Compact mode to show on mobile (or with a toggle).

Geometrically:
Considering that these effectively almost double creator payouts, I think the ~10 pixels really doesn’t matter all that much on mobile. The ethicalads on mobile were a very similar size

HyperSoop:
wait did you just merge it with the name "adrinth"? i thought this was supposed to be a haha funny meme placeholder

Geometrically:
> so, are those the in-house ads? Those are a huge additon and should be announced.

This isn't the full in house ads plan we have yet. We're just starting with one advertiser and building up from there :)

Aces-and-Jacks:
> why's it labelled "adrinth" it's just a meme though

I think adrinth is a pretty good name for it

HyperSoop:
it's a meme, a funny one at that, but still just a meme. Plus, the in-house ads are not a separate Rinth service from Modrinth, thought such a name implies they are.

PR #748, opened 2022-11-17
Remove bottom padding on <details>' last element

EnnuiL:
This removes the empty space that are inside `<details>`'s end:
Crop of Ok Zoomer's Presets spoiler, with the fix applied

The below has been addressed on the refined version of the fix
<details>
<summary>Outdated note</summary>

Do note that if the `<details>` is improperly formatted, then this fix will slightly break it even further, although considered that it's already broken, I don't believe that this outweighs the benefits of the fix;
Crop of Ok Zoomer's Presets spoiler edited in order to be badly formatted, with the fix applied, making the padding slightly wonky
</details>

(dev note: why is node.js v16 required when v18 is already LTS ;-;)

### DIFF
--- a/assets/styles/components.scss
+++ b/assets/styles/components.scss
@@ -113,7 +113,14 @@
     border-bottom: 1px solid var(--color-header-underline);
   }
 
-  h1, h2, h3, h4, h5, h6, li, p {
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6,
+  li,
+  p {
     word-wrap: break-word;
     overflow-wrap: anywhere;
   }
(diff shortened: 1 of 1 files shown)


triphora (dismissed):
Please revert all of the unnecessary whitespace changes.

EnnuiL:
wait, what in the world did my IDE do

EnnuiL:
oh, of course everything was breaking, the repository's VSCode settings has auto-formatting on...

Anyway, I've went ahead and improved the fix, addressing the potential con of this PR

Prospector (approved):

triphora (approved):

PR #753, opened 2022-11-19
Migrate to lucide-vue for icons and update copying info

triphora:


Prospector:
I started making a list of icons that I felt looked notably worse than before, but I'm not sure it's worth actually fixing these because just using all lucide would really be quite helpful. The only one I think definitely needs to be changed is the money icon, it's just not recognizable

Note | Current |  Lucide
:-:|-|:-:
Download icon just looks weird and poorly proportioned | image  |  image
A bill at this size just looks like a rectangle and not money | image | image
The narrower shoulders on the Account is just less distinct and is just marginally worse | image | image
Project cards use an "Edit" icon instead of an "Updated" icon | image | image
Search icon looks unbalanced and poorly proportioned | image | image
New caret doesn't match the font weight as nicely | image | image

pages/dashboard.vue
```diff
@@ -30,14 +30,16 @@
 </template>
 
 <script>
+import {
+  LayoutDashboard as DashboardIcon,
+  // BarChart3 as ChartIcon,
+  Banknote as CurrencyIcon,
```
Prospector, line 36: Use dollar sign instead of bank note

pages/settings.vue
```diff
@@ -33,14 +33,16 @@
 </template>
 
 <script>
+import {
+  Paintbrush as PaintbrushIcon,
+  User as UserIcon,
+  Heart as HeartIcon,
+  Banknote as CurrencyIcon,
```
Prospector, line 40: Use dollar sign instead of bank note

Prospector (changes requested):

Prospector:
I actually just noticed this one, without the circle it doesn't even look like the GitHub logo:

image

triphora:
Closing in favour of #760

PR #755, opened 2022-11-20
Use safe insets on bottom of screen

Prospector:
Should fix the overlay of the iPhone navigation gestures. I cannot test it though as I do not own such a device.


haykam821 (approved):
image

triphora (approved):

PR #756, opened 2022-11-20
Improve badges and work on a more unified color scheme for random colors

Prospector:
Examples of the new badges and color scheme in action:

image
image

### DIFF
--- a/assets/images/utils/archive.svg
+++ b/assets/images/utils/archive.svg
@@ -1,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+  <rect x="2" y="4" width="20" height="5" rx="2"></rect>
+  <path d="M4 9v9a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V9"></path>
+  <path d="M10 13h4"></path>
+
+</svg>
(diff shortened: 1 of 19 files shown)


triphora (approved):

Andre601:
Quickly providing my own feedback on this...
How would this be in regards to colour blind people? Those suffering a red-green blindness may have a difficult time seeing a difference between the release and the (I assume) alpha releases. Perhaps the changelog should in a future update have some more unique symbols to indicate the type of release?

Of course, I'm not one having a color blindness, so I could be wrong here on if that is having a negative impact at all... But it should be considered for the future.

Jamalam360:
Oh hello there. I'm quite serverely red-green colourblind. I'll preface this by saying the condition is very variable from person to person, even with the same type (red-green etc.), but from the screenshot those colours are definitely defined enough for me to see the difference quite clearly.

Geometrically (approved):

Andre601:
> Oh hello there. I'm quite serverely red-green colourblind. I'll preface this by saying the condition is very variable from person to person, even with the same type (red-green etc.), but from the screenshot those colours are definitely defined enough for me to see the difference quite clearly.

Thanks for the clarification.
I quickly tested around using a a11y add-on to see and from the testing, people with Achromatomaly would have the strongest impact (See screenshot)

But as you said does it all depend on the person, since not everyone has the same level of impact on it.
Just think that a symbol of sorts could still help in the future... Tho then we would need to make it clear what the icon means. :/
Maybe a title on hover saying the type (If that is doable)?

Screenshot as mentioned:
grafik

Prospector:
At least for me, even in that screenshot the brightness of the color is distinct enough to distinguish each release channel for me. I do think adding an icon is a good idea though and maybe something we would consider in the future. In fact, this PR replaces pure color circles with icons in other places with the badges

PR #762, opened 2022-11-21
Add project redir + fix version 404

Geometrically:


triphora (approved):
Tested on the deploy and works fine

PR #763, opened 2022-11-21
Corresponding frontend PR to modrinth/labrinth#449

### LINKED ISSUE #593: Selecting a license deletes already defined URL
### Environment

Website

### Describe the bug

When you set a URL for the license and then select the License itself is the URL field overridden to either the CDN url, or simply emptied for CUSTOM.

### Steps To Reproduce

1. Create a new project
2. Scroll down to "License"
3. Paste a URL in the "License URL" field
4. Select a URL from the dropdown

### Expected behavior

The URL field shouldn't be overriden if you pasted a URL without having a license selected yet.
At least for `CUSTOM` should this be the case.

### Additional context

_No response_

### LINKED ISSUE #597: License URL reverts back to default when editing project
### Environment

Website

### Describe the bug

When going to the project edit page the license URL always reverts back to the default one on the modrinth CDN, rather than keeping the custom URL set on previous visits to the site.

### Steps To Reproduce

- Set and save a custom license URL for a project
- Re-open the project edit page and scroll down to see the default value appear again
  - Upon saving this will change the URL used on the mod page, which is unwanted

### Expected behavior

The link to my license should be kept.

### Additional context

This might be related to #593, however it also happens without changing the license that is used.

### LINKED ISSUE #768: Add OSLv3 license support

_No response_

OSLv3 is a license that forces forks to be open source, like GPLv3, and doesn't allow server owners to use the mod without giving the users of the server the mod's source. (Letting users know the server owner didn't make the mod themselves)

Using the CUSTOM one; yet I prefer for modrinth to show my actual license.

### Additional context

_No response_

triphora:
Redo of #688
Fixes #597
Fixes #593
Fixes #768


Prospector (approved):

assets/styles/components.scss
```diff
@@ -1042,8 +1043,12 @@ h1 {
...
 
-  input + *, .input-group + * {
-    margin-block-start: var(--spacing-card-md);
+  >,
+  .extend-styling> {
+    input + *,
+    .input-group + * {
+      margin-block-start: var(--spacing-card-md);
+    }
   }
```
Geometrically, line 1052: Kind of unrelated to this PR, but I really dislike this styling block. It's not super readable and there's probably better ways to do this

pages/_type/_id.vue
```diff
@@ -471,15 +477,29 @@
...
+                  {{ project.license.id.replaceAll('LicenseRef-', '') }}
+                </a>
+                <a
+                  v-else-if="
+                    project.license.id === 'LicenseRef-All-Rights-Reserved'
+                  "
+                  class="text-link"
+                  @click="getLicenseData()"
+                >
+                  All Rights Reserved
```
Geometrically, line 495: This looks pretty weird in the UI. Could it just be "ARR"?

pages/_type/_id.vue
```diff
@@ -471,15 +477,29 @@
...
+                </a>
+                <span v-else-if="project.license.id.includes('LicenseRef')">
+                  {{ project.license.id.substring(10).replaceAll('-', ' ') }}
+                </span>
+                <a v-else class="text-link" @click="getLicenseData()">
                   {{ project.license.id }}
                 </a>
```
Geometrically, line 502: Hmm, I don't really like the handling of license names here. It makes a lot more sense to process this in `asyncData` or something rather than in the template.

pages/_type/_id.vue
```diff
@@ -812,15 +832,29 @@
...
+                  {{ project.license.id.replaceAll('LicenseRef-', '') }}
+                </a>
+                <a
+                  v-else-if="
+                    project.license.id === 'LicenseRef-All-Rights-Reserved'
+                  "
+                  class="text-link"
+                  @click="getLicenseData()"
+                >
+                  All Rights Reserved
```
Geometrically, line 850: see other comments about this section in the other side of the page, also "All Rights Reserved" => ARR

pages/search.vue
```diff
@@ -195,17 +195,12 @@
...
-            :options="$tag.licenses.map((x) => x.short.toUpperCase())"
-            :multiple="true"
-            :searchable="true"
-            :close-on-select="false"
-            :show-labels="false"
-            :allow-empty="true"
+          <h3 class="sidebar-menu-heading">Open source</h3>
+          <Checkbox
+            v-model="onlyOpenSource"
+            label="Only include open source projects"
```
Geometrically, line 201: The label is a little too verbose IMO. "Open source" should suffice. The other category filters aren't phrased this weird so it seems a little out of place.

pages/_type/_id.vue
```diff
@@ -1,5 +1,11 @@
 <template>
   <div>
+    <ModalLicense
+      ref="modal_license"
+      :license-name="project.license.name ? project.license.name : 'License'"
+      :license-id="project.license.id"
+      :license-text="licenseText"
+    />
```
Geometrically, line 8: Couldn't this just use a raw `Modal`? doesn't seem to be used on multiple pages

pages/_type/_id.vue
```diff
@@ -1193,6 +1230,17 @@ export default {
...
     },
+    async getLicenseData() {
+      try {
+        const text = await this.$axios.get(
+          `tag/license/${this.project.license.id}`
+        )
+        this.licenseText = text.data.body
+        this.$refs.modal_license.show()
+      } catch {
+        this.licenseText = 'License text could not be retrieved.'
```
Geometrically, line 1241: Shouldn't the modal show even though the request fails? Just so they can see this message

pages/_type/_id/edit.vue
```diff
@@ -382,55 +379,103 @@
...
-            placeholder="License URL"
+            :placeholder="nonSpdxLicense ? 'License name' : 'SPDX identifier'"
             :class="{
-              'known-error': newProject.license_url === '' && showKnownErrors,
+              'known-error': license.short === '' && showKnownErrors,
             }"
             :disabled="
               (currentMember.permissions & EDIT_DETAILS) !== EDIT_DETAILS
             "
           />
```
Geometrically, line 467: In the UI it is not very intuitive that you don't have to set a license URL anymore

pages/_type/_id/edit.vue
```diff
@@ -382,55 +379,103 @@
...
+            <a
+              href="https://blog.modrinth.com/licensing-guide/"
+              target="_blank"
+              rel="noopener noreferrer"
+              class="text-link"
+            >
+              licensing guide</a
+            >
+            for more information.
+          </span>
```
Geometrically, line 417: Breaks aren't rendering for me for some reason. It looks like one giant text block. I think you may need to put double breaks so there's space in b/w paragraphs. would look a lot better

Geometrically (changes requested):
Did a brief review, will look more into edit page on second glance

pages/search.vue
triphora: The main issue is that, with that phrasing, it will imply that open source projects are excluded if it's unselected by default, whereas this phrasing makes it explicit that it's excluding non-open-source projects. Would "Open source only" be good?

Geometrically: Does it? For the categories it doesn't seem that way. "Open source only" is good too but I think either works. Like we don't have "Adventure only" or "Fabric only".

pages/_type/_id.vue
triphora: Cleaned up this section

PR #765, opened 2022-11-22
Flameanvil sync

Geometrically:


triphora:
Going to close this since the relevant backend work needs to be redone, and even when this is done, it will not use this branch due to how many conflicts there are

PR #766, opened 2022-11-22
Re-word "my profile" to "your profile"/"this profile"

sisby-folk:
We're like half asleep and totally open to actual design people picking different wording for these strings, but I'm not super familiar with who's on the designs team and how to get in contact with them so this seemed like a good direct way.

Anyway this pr does [drumroll]

image     
becomes 
image     

image
becomes
image

not super fussed on the second string seeing as we'll never see it but we included it for comprehensiveness

that's it,

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -87,7 +87,7 @@
                   >
                     <div class="title profile-link">
                       <div class="username">@{{ $auth.user.username }}</div>
-                      <div class="prompt">Go to your profile</div>
+                      <div class="prompt">Visit your profile</div>
                     </div>
                   </NuxtLink>
                   <hr class="divider" />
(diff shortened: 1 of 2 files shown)


Prospector:
I was meaning to change it to "Visit your profile" to be more in line with other button links. Could you change it to that?

sisby-folk:
🐉 Done!

layouts/default.vue
```diff
@@ -228,7 +228,7 @@
               />
               <div class="profile-link">
                 <div class="username">@{{ $auth.user.username }}</div>
-                <div class="prompt">Go to my profile</div>
+                <div class="prompt">Go to your profile</div>
```
Prospector, line 231: just here too

Prospector (changes requested):

layouts/default.vue
sisby-folk: thank you, in our hubris we made the change on a phone web browser 😌

triphora (approved):

Prospector (approved):

PR #772, opened 2022-11-25
Restore transparency in project icons

Gaming32:


Rafii2198:
Instead of removing shadow completely, I would replace it with drop-shadow, so it would look like this:
obraz

*Note that these values are not original values and I just made them to see how it looks and of course is just my opinion*

Prospector:
This change was intentional

Gaming32:
> Instead of removing shadow completely, I would replace it with drop-shadow, so it would look like this: obraz
> 
> _Note that these values are not original values and I just made them to see how it looks and of course is just my opinion_

That looks good.

Gaming32:
> This change was intentional

Doesn't make it necessarily the best change. I got quite a few people on Discord saying they prefer the changes in this PR.

intergrav:
I agree. I feel like the icon background isn't that necessary

Gaming32:
Discussion moved to https://discord.com/channels/734077874708938864/1045786101673111553 (Modrinth Discord)

PR #773, opened 2022-11-25
Add gallery view to search pages

Prospector:
image


Minenash:
Can you also provide what it would look like for protects without a featured gallery image, which would be very common in the mod section?

assets/styles/components.scss
```diff
@@ -479,6 +483,32 @@ tr.button-transparent {
   }
 }
 
+
+
```
triphora, line 487: excessive number of newlines

triphora (approved):
That was literally my only comment going through the code. Good job

PR #776, opened 2022-11-26
Fix notifications fetching

Brawaru:
In actions of user Vuex store, calling an action to fetch notifications would fetch them correctly, but store them in place of user follows, which resulted in funkiness when switching from notifications to follows page (because Vue would render project cards for notifications data before the actual follows were fetched correctly).


triphora (approved):
Thanks!

PR #783, opened 2022-11-28
Launcher Integration

trash-pandy:
A partial implementation for issue https://github.com/modrinth/knossos/issues/380.
Currently adds a button next to all downloads that launches a protocol handler intended to be received by an application on the user's computer that is configured to handle it. The protocol handler currently takes the form `modrinth://<project-type>/<version-id>`, where `<project-type>` is the type of project being requested and `<version-id>` is the version of the item selected to be downloaded. There are settings for disabling this feature entirely as well as switching the "Download" and "Install with Launcher" buttons.
Implemented a first-run dialogue which guides the user through their first time using the feature, as well as allowing them to hide the feature entirely. Includes links to the documentation section on installing modpacks, but should probably have its own section on installing launchers specifically with this feature.

### DIFF
--- a/components/ui/ModalDidThatWork.vue
+++ b/components/ui/ModalDidThatWork.vue
@@ -1,0 +1,110 @@
+<template>
+  <Modal ref="modal" header="Did that work?">
+    <div class="modal-didthatwork">
+      <span>
+        Did your launcher open? It might take a moment, or you might need to
+        accept a prompt. If it didn't open, you may need to update or
+        <a
+          href="https://docs.modrinth.com/docs/modpacks/playing_modpacks/#how-do-i-play-a-modpack"
+        >
+          install a compatible launcher.
+        </a>
+      </span>
+      <section class="universal-card">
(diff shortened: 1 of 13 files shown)


components/ui/ModalDidThatWork.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+  <Modal ref="modal" header="Did that work?">
+    <div class="modal-didthatwork">
+      <span>
+        Did your launcher open? It might take a moment, or you might need to
+        accept a prompt. If it didn't open, you may need to update or
+        <a
+          href="https://docs.modrinth.com/docs/modpacks/playing_modpacks/#how-do-i-play-a-modpack"
```
jadelily18, line 8: this should probably open a new tab but it looks really really good!!

Geometrically:
we'll be doing this another way in a future PR. thanks for the PR though!

Minenash:
If you're curious:
https://floss.social/@modrinth/110607822611108529

PR #812, opened 2022-12-18
Change 'changelog-item' display to block from flex.

frikinjay:
Change display to block from flex to fix content overflow on the changelog page.


frikinjay:
nvm this doesnt fix the issue. No idea why it worked in dev, it shouldnt have.

triphora (approved):

PR #897, opened 2023-01-11
Remove numbers at the start of slugs when converting to mod ID

### LINKED ISSUE #845: Packaged datapack invalid if name start by a number
### Describe the bug

When your creates a packaged datapack, when the project name starts with a number, the id generated is invalid and raise a error when your try to play with the packaged datapack.

### Steps to reproduce

1. Create a new projet where the name begins with a number, like 123-test
2. Upload a datapack
3. Packaged the datapack to a mod
4. Download and install the packaged datapack to your minecraft mods folders
5. Lauch Minecraft

### Expected behavior

Once installed in the minecraft mods folders, Minecraft should be able to run without any problem with the packaged datapack loaded.

### Additional context

After some test, it seems the id of a mod must start only with a-z characters.
Also, I've only tested with Quilt, but I wouldn't be surprised if Fabric or Forge raise the same error.

Prospector:
Removes any leading numbers from the mod ID that is generated from the project's slug for data pack -> mod conversion

Fixes #845

### DIFF
--- a/plugins/fileUtils.js
+++ b/plugins/fileUtils.js
@@ -363,6 +363,7 @@ export const createDataPackVersion = async function (
   const newSlug = `${project.slug
     .replace('-', '_')
     .replace(/\W/g, '')
+    .replace(/^(\d+)(\D+)$/g, '$2')
     .substring(0, 63)}_mr`
 
   const iconPath = `${project.slug}_pack.png`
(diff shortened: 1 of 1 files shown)


triphora (approved):

Geometrically (changes requested):
This doesn't seem to work with number only slugs- for those I think we should just replace with Project ID

Prospector:
Tbh, not sure number-only slugs should even be allowed (how can your project be just called a number?) but I added that

Geometrically (approved):

PR #900, opened 2023-01-11
Fix a couple of misc. page errors I noticed

Prospector:
One with the notifications page crashing and the other with some console spam on project pages.


triphora (approved):

Geometrically (changes requested):
PR is good to merge but could you add `status_update` to NOTIFICATION_TYPES? I think right now they don't show up in the frontend.

Prospector:
Added

Geometrically (approved):

PR #902, opened 2023-01-11
Add release channel filter to versions pages

### LINKED ISSUE #674: Ability to filter versions by release channel

I publish development builds continuously on the alpha channel and release builds manually on the release channel. As a result, when I visit the releases or changelog pages things get messy as my many alpha-channel builds are mixed in with the release ones.
<details>
<summary>Screenshots</summary>
<br>

image
image
</details>

The ability to filter by release channel on the changelogs and versions tabs of projects. It'd be nice if it gave you a little set of three checkboxes (alpha, beta, release) and they'd only show up if there are builds in the relevant channels.

Separate tabs for each release channel, perhaps? I think the filtering solution is a bit of a better one.

### Additional context

The current version filter options are tailored more towards mods. With plugins, filtering by Minecraft version is fairly irrelevant due to the nature of the Bukkit API platform and since most plugins are compatible with downstream "loaders", that's also not particularly helpful. There's not really much you can do about that, it's just one of those things I guess.

Prospector:
Closes #674

### DIFF
--- a/components/ui/VersionFilterControl.vue
+++ b/components/ui/VersionFilterControl.vue
@@ -39,6 +39,24 @@
       placeholder="Filter versions..."
       @input="updateVersionFilters()"
     ></Multiselect>
+    <Multiselect
+      v-if="getValidChannels().length > 1"
+      v-model="selectedChannel"
+      :options="getValidChannels()"
+      :custom-label="
+        (x) => (x === 'any' ? 'Any channel' : $capitalizeString(x))
+      "
+      :multiple="false"
+      :searchable="false"
+      :show-no-results="false"
+      :close-on-select="true"
+      :clear-search-on-select="false"
+      :show-labels="false"
+      :allow-empty="true"
+      class="release-channel"
+      placeholder="Filter channels..."
+      @input="updateVersionFilters()"
+    ></Multiselect>
     <Checkbox
       v-if="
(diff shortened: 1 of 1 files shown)


components/ui/VersionFilterControl.vue
```diff
@@ -39,6 +39,24 @@
...
+      :searchable="false"
+      :show-no-results="false"
+      :close-on-select="true"
+      :clear-search-on-select="false"
+      :show-labels="false"
+      :allow-empty="true"
+      class="release-channel"
+      placeholder="Filter channels..."
+      @input="updateVersionFilters()"
+    ></Multiselect>
```
Geometrically, line 59: How come it's a single select? For consistency + additional functionality I feel like it would be better if you could choose multiple and the ones you want.

Geometrically (changes requested):

components/ui/VersionFilterControl.vue
Prospector: It just seemed kinda weird as a single select and not really how most would want to use it as it made it harder to switch from one channel to another (say, from showing just Releases to just Alphas), but I can change it if you think I should

Prospector: Changed

Geometrically (approved):

PR #905, opened 2023-01-11
Fix custom license erroring with spaces

Prospector:
Pretty sure this was the intended functionality of this, it'll replace spaces in custom license names with `-` for the ID


triphora (approved):
Indeed it was

PR #928, opened 2023-01-18
Remove .zip from possible modpack versions

thewilloftheshadow:
This format isn't allowed in Labrinth, and shouldn't be shown as a option in the file picker

https://github.com/modrinth/labrinth/blob/df3aeed2910cf11d39fe0dbe1a05f411e0541397/src/validate/modpack.rs#L15

### DIFF
--- a/plugins/fileUtils.js
+++ b/plugins/fileUtils.js
@@ -41,7 +41,7 @@ export const acceptFileFromProjectType = (projectType) => {
     case 'datapack':
       return '.zip,application/zip'
     case 'modpack':
-      return '.mrpack,application/x-modrinth-modpack+zip,application/zip'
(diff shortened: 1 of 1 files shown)


Minenash:
I know for mods, zips are often used for sources, is there a like wise reason to allow zips as secondary files for modpacks?

triphora:
I don't think there is @Minenash

triphora (commented):
Are you sure this still recognizes mrpack files on systems/browsers which base file pickers on MIME types? We can't expect every application to be using `application/x-modrinth-modpack+zip`

Geometrically:
Closing as most file pickers would not recognize the MRPack file type, and just have it as a ZIP.

PR #960, opened 2023-01-30
 Set layout display to flow-root (fixes #958)

NeutronicMC:
This prevents pages from overflowing by default. Previously, margins from children of containers with `.layout` would apply outside of the body instead of within. (Fixes #958)

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -500,7 +500,7 @@ export default defineNuxtComponent({
 .layout {
   min-height: 100vh;
   background-color: var(--color-bg);
-  display: block;
+  display: flow-root;
 
   @media screen and (min-width: 1024px) {
     min-height: calc(100vh - var(--spacing-card-bg));


triphora (commented):
Doesn't seem to fix the linked issue on my end.

NeutronicMC:
> Doesn't seem to fix the linked issue on my end.

What browser do you use?

triphora:
Tested it on Firefox

MagnusHJensen:
I see the issue.
I think it's more a matter of the page, being a fixed `min-height: 100vh` that forces a scrollable area, if the content is moved-around.

If disabled the page is no longer scrollable.
It's on the `.layout` styling.

NeutronicMC:
> I see the issue.
> 
> I think it's more a matter of the page, being a fixed `min-height: 100vh` that forces a scrollable area.
> 
> 
> 
> If disabled the page is no longer scrollable.
> 
> It's on the `.layout` styling.

Yep, that sounds right to me. Not sure how I thought the original PR was working, I must've been up too late at night because I haven't been able to reproduce it since 😅

I'll make those changes tomorrow as well

PR #966, opened 2023-01-31
Process spoiler tags when pasted from CurseForge (Closes #310)

### LINKED ISSUE #310: Autoconvert curseforge spoiler tags into details tags in markdown

When maintaining a mod on both modrinth and curseforge, it would be super helpful to be able to copy/paste the curseforge description into modrinth without needing to change anything about the markdown. However, curseforge uses <div class="spoiler"> to create collapsible sections while modrinth requires <details>. 

A solution would be to have modrinth take the curseforge markdown with spoilers and autoconvert it to details tags for the poster. Example of some curseforge markdown it might have to handle:
```
<p><span style="font-size: 14px;"><strong>Enchantments&nbsp;</strong>- 1 new enchantment</span></p>
<div class="spoiler">
<p><strong>&nbsp;Comb Cutter</strong></p>
<div class="spoiler">
<p>-Can be applied to Swords and Shears</p>
<p>&nbsp;</p>
<p>-Only 1 level but will make the tool mine all comb based blocks super fast! It will also mine wax, hive, and nest blocks slightly faster.</p>
<p>&nbsp;</p>
<p>-Can be used to double the Honeycomb output from Beehives and Bee Nests.</p>
</div>
</div>
```

which looks like this on curseforge:
image

On modrinth, it should read this markdown something like this
```

### Enchantments - 1 new enchantment

<details>

**Comb Cutter**

<details>

-Can be applied to Swords and Shears

&nbsp;

-Only 1 level but will make the tool mine all comb based blocks super fast! It will also mine wax, hive, and nest blocks slightly faster.

&nbsp;

-Can be used to double the Honeycomb output from Beehives and Bee Nests.

</details>

</details>
```

Which will look like this:
image

The closer the converted markdown looks to the curseforge's result, the better and easier it will be for modders to switch to modrinth and keep their descriptions up to date. Just a good suggestion

_No response_

### Additional context

_No response_

NeutronicMC:
Closes #310

Just did this for a fun challenge, no big deal if it doesn't fit within project scope considering how long ago the issue was made.

Basically, this checks any text pasted into the project description box for `<div class="spoiler">` tags and replaces them with `<description>` tags. My code parses the raw text into a DOM object in order to properly pair tags and support nested tags. It won't interfere with paste events that don't contain a spoiler div.

Overall, this could be a nice quality-of-life change for authors that use both Modrinth and CurseForge, or that want to move from CurseForge to Modrinth.

### DIFF
--- a/pages/[type]/[id]/settings/description.vue
+++ b/pages/[type]/[id]/settings/description.vue
@@ -27,6 +27,10 @@
           id="project-description"
           v-model="description"
           :disabled="(currentMember.permissions & EDIT_BODY) !== EDIT_BODY"
+          @input="processInput"
+          @keyup="updateCursorPosition"
+          @click="updateCursorPosition"
+          @select="updateCursorPosition"
         />
       </div>
       <div
@@ -95,6 +99,8 @@ export default defineNuxtComponent({
   data() {
     return {
       description: this.project.body,
+      original: '',
+      cursorPosition: 0,
       bodyViewMode: 'source',
     }
   },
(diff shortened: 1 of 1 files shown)


MagnusHJensen:
While I agree on this, and the better the _cross-compatibility_ is, it makes the site easier to use for mod devs.

However I don't think this is the right approach, I think of 2 issues.

1. This is unexpected to the user. The reason for this is, that while I might expect it to also do some sort of "spoiler", I would not expect it to straight up change my plain text. Especially not without user-consent. I think it would be a better way to either edit how <div class="spoiler"> is rendered in the markdown preview, instead of changing the text, or prompt the user if the text contains certain things, which can be transformed if they want them transformed.

2. I don't think this functionality should live in the description page itself, I think this should be in a utility file somewhere, either in the markdown plugin itself, since it strictly relates to markdown, or some text-util file. This also allows for other places where markdown is used, to have this transform functionality, rather than only the description.

I do like the idea though, and it works just fine!

NeutronicMC:
Yeah, creating a dismissible banner would be a much better UX. I never really iterated on this concept because of the implications of supporting something like this (the logistics of supporting other formats would be messy). Regardless, if you'd still like me to implement a banner and move it to a utility file that'd be great as well.

PR #980, opened 2023-02-02
Show supported versions on project page, fixes #463

### LINKED ISSUE #463: Show supported versions on mod page

I think the project could show the supported versions
One issue I find when using another mod platform is that while it shows the supported version, it shows of the last release, and with this, I need to check the files to see if the mod supports a newer version, or an older one, it is great to be able to just quickly go on the page an see if the mod supports the game version you want, checking the "Featured versions" isn't really practical, specially when the mod supports many versions, and so the list becomes quite large

The supported could be shown under update and create date, being on a location where it is easy to spot

I made an example image of how it could look like
image

An alternative I have considered is show supported versions above the description
image

### Additional context

_No response_

CodeF53:
Fixes #463
image
image

### DIFF
--- a/pages/_type/_id.vue
+++ b/pages/_type/_id.vue
@@ -237,6 +237,12 @@
                 }}</span>
               </div>
             </div>
+            <div class="versions">
+              <span class="label">Supports</span>
+              <span class="value">
+                {{ $formatVersion(project.game_versions) }}
+              </span>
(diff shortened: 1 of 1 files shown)


Geometrically:
we'll be doing this in another way in a future update. thanks for the PR though!

PR #1001, opened 2023-02-18
fix(xss): allow for any query parameters from youtube embed

MagnusHJensen:
**Description:**
The previous regex ONLY allowed for youtube embeds to have the query param `autoplay`, meaning when embedding youtube videos, people would have to manually sort through their src url to remove unwanted properties. (This is also not wanted, since the `start` is a very useful prop for these types of videos, if the real tutorial/showcase begins 2 minutes in.

The following regex works for any query parameters. 
`^https?:\/\/(www\.)?youtube(-nocookie)?\.com\/embed\/[a-zA-Z0-9_-]{11}(\?(&?\w+=(\w|\d)+)+)?$`
(This setup then works as a default blacklist approach, where any non-wanted query parameters should be added to the remove section).
Simply due to the nature of regex, that requires the position of strings to match, or just matching 3 parameters would end up in a very clonky `a,b,c | a,c,b | c,a,b | etc... (9 options total)`

**Media:**
With new regex:

https://user-images.githubusercontent.com/32039051/219852221-981930aa-6160-484e-a047-7ca6300a4f62.mov

Old regex:
image

Closes: #362


pandaninjas:
This new regex is ReDOSable and could even cause the JS to throw a `Uncaught InternalError: too much recursion`.

You can see this in action with the string `http://youtube.com/embed/-dtieshosdb?0=o0000000000000000000000000000000`.  It takes JS a long time to execute, and then it throws `Uncaught InternalError: too much recursion`.

plugins/xss.js
```diff
@@ -30,7 +30,7 @@ const options = {
       const allowedSources = [
         {
           regex:
-            /^https?:\/\/(www\.)?youtube(-nocookie)?\.com\/embed\/[a-zA-Z0-9_-]{11}(\?&autoplay=[0-1]{1})?$/,
+            /^https?:\/\/(www\.)?youtube(-nocookie)?\.com\/embed\/[a-zA-Z0-9_-]{11}(\?(&?\w+=(\w|\d)+)+)?$/,
```
pandaninjas, line 33:
```suggestion
            /^https?:\/\/(www\.)?youtube(-nocookie)?\.com\/embed\/[a-zA-Z0-9_-]{11}((&|\?)\w+=\w+)*$/,
```
I've tried to fix the ReDOS here

pandaninjas (commented):
In addition, the `remove` also needs to be fixed, because one can do `?autoplay=1` to get the same effect. Before trying that would *fail* the regex, due to the `\?&`

MagnusHJensen:
Closed in favor of #1064

PR #1012, opened 2023-02-24
Fixes navbar transition animation.

ZachBaird:


components/ui/NavRow.vue
```diff
@@ -10,6 +10,7 @@
     >
       <span>{{ link.label }}</span>
     </NuxtLink>
+    <div class="nav-indicator" :style="{ left: positionToMove, width: sliderWidth }"></div>
```
Geometrically, line 13: should add `aria-hidden="true"` to hide it from screen readers

Geometrically (changes requested):
Looks good but the animation needs some changes:
- @Prospector might have some additional feedback but the animation when the line is "hidden" (ex: moving from modpacks to the home page). It feels weird that it goes all the way across the screen
- Animation feels a little slow (maybe speed it up a bit)
- Animation doesn't work vertically (for mobile)

pages/[type]/[id].vue
```diff
@@ -290,10 +290,9 @@
...
           </div>
           <div
-            v-highlightjs
             class="markdown-body"
             v-html="
-              $xss(
+              renderString(
                 $md.render(
                   project.moderator_message.body
                     ? project.moderator_message.body
```
Comment (Geometrically) [inline pages/[type]/[id].vue:298]: Look at /index.vue (in this directory) for an example on this func.

It should just be renderString(project.moderator_message.body ? ... (whatever else). Also make sure to import the function (you can also see index.vue for the file to import from)

Geometrically (changes requested):
Looks good besides below

Geometrically (approved):

PR #1013, opened 2023-02-25
allow negative filter on categories

ItsiAdam:
This is a lazy implementation of negative category filtering. Adds a third state to Checkbox which doesn't break existing behaviour. I'll happily adjust to satisfy any requirements/feedback.

I think this just needs URL params to be implemented and some code comments. 

image

### DIFF
--- a/components/ui/Checkbox.vue
+++ b/components/ui/Checkbox.vue
@@ -9,11 +9,22 @@
       class="checkbox"
       role="checkbox"
       :disabled="disabled"
-      :class="{ checked: value, collapsing: collapsingToggleStyle }"
+      :class="{
+        checked: value == 1,
+        negativeChecked: value == 2,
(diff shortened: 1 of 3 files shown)


falseresync:
What about checkboxes in other parts of the website? For example, in project settings? This has to be a separate component (TristateCheckbox) or a prop to choose between 2 and 3 state modes

I personally don’t see much use in this

Minenash:
> I personally don’t see much use in this
Do you mean of the feature itself? Because it's *definitely* useful

triphora (changes requested):
Just echoing what false mentioned here:
> What about checkboxes in other parts of the website? For example, in project settings? This has to be a separate component (TristateCheckbox) or a prop to choose between 2 and 3 state modes

triphora:
Will close soon if reviews aren't addressed

PR #1022, opened 2023-03-01
Nuxt 3 Lazy Loading Search

ZachBaird:
~~This PR modifies the `useBaseFetch` composable to use Nuxt 3's `useLazyFetch` composable.~~

~~Near as I can tell, `useLazyFetch` requires the method to be used either at the top-level of a `<script setup>` or in a lifecycle method. As of creating this PR, I haven't had any success getting it to work inside onMounted.  Consequently, I've gone and converted the search page into a `<script setup>` type of component so I could work with things a bit easier. Due to a lot more code being top-level now, I do think we need to move, some of the code in the search page to helpers, composables, and components we render in the page - wherever it makes sense. The page itself feels a bit bloated. Since this branch was meant to fix the stuttering and get the search page to be more fluid, such refactoring was out-of-scope for now.~~

~~With the above changes, the search page's request for data now lazy loads. The UI will load cleanly without stuttering while the data fetches in the background. A "Loading..." string populates the UI until the response comes in.~~

Search page uses useLazyFetch with transform.

**Notes**: 

- Loading circle styling does not work on grid view

### DIFF
--- a/pages/search/[searchProjectType].vue
+++ b/pages/search/[searchProjectType].vue
@@ -312,7 +312,7 @@
           role="list"
           aria-label="Search results"
         >
-          <LogoAnimated v-if="searchLoading && results?.hits?.length === 0" />
+          <LogoAnimated v-if="searchLoading && (!results || results?.hits?.length === 0)" />
           <ProjectCard
             v-for="result in results?.hits"
             v-else
--- a/pages/search/[searchProjectType].vue
+++ b/pages/search/[searchProjectType].vue
@@ -312,8 +312,14 @@
           role="list"
           aria-label="Search results"
         >
+          <div v-if="searchLoading" class="orbit-spinner">
+            <div class="orbit"></div>
+            <div class="orbit"></div>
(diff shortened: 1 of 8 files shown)


pages/search/[searchProjectType].vue
```diff
@@ -312,8 +312,14 @@
           role="list"
           aria-label="Search results"
         >
+          <div v-if="searchLoading && results?.hits?.length === 0" class="orbit-spinner">
+            <div class="orbit"></div>
+            <div class="orbit"></div>
+            <div class="orbit"></div>
+          </div>
```
Comment (Geometrically) [inline pages/search/[searchProjectType].vue:319]: Could we use the old Modrinth animated logo animation here? this looks a little out of place

Geometrically (changes requested):
Mostly good, but one minor issue: when switching between project types (ex: from plugins -> data packs), there is a short period where it shows the old project results. this could be fixed by adding a watcher for the project type which sets the results to undefined when the project type changes.

Geometrically (approved):

PR #1038, opened 2023-03-11
fix(project): update follower count client side

MagnusHJensen:
**Description:**
Ensures the project followers on the client side is in-sync with the actions taken.

**Media:**

https://user-images.githubusercontent.com/32039051/224480515-ef39e6ba-55ac-455f-885e-56efd42607cd.mov

Closes: #437

### DIFF
--- a/composables/user.js
+++ b/composables/user.js
@@ -82,6 +82,7 @@ export const userFollowProject = async (project) => {
   const user = (await useUser()).value
 
   user.follows = user.follows.concat(project)
+  project.followers++
 
   setTimeout(() => {
     useBaseFetch(`project/${project.id}/follow`, {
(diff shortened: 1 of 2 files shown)


triphora (commented):
Wouldn't it be easier to add it to user(un)followProject in composables/user.js?

MagnusHJensen:
Maybe, I just Saw the composables files as they should only have Logic relater to the name/file so user in this case, where incrementing and decrementing is not user related but more page specific in My mind.
I can easily change it if I misunderstood how they are separated :D

triphora:
tbh, I don't know how it's "supposed to be" separated. In any event, incrementing and decrementing is going to be wanted no matter which page it's on.

MagnusHJensen:
Moved to user composable

triphora (approved):

PR #1064, opened 2023-03-26
 fix(xss): allow any query parameters in youtube embed, ported from #1001

pandaninjas:
**Description:**
(This has been copied from #1001 verbatim)
The previous regex ONLY allowed for youtube embeds to have the query param autoplay, meaning when embedding youtube videos, people would have to manually sort through their src url to remove unwanted properties. (This is also not wanted, since the start is a very useful prop for these types of videos, if the real tutorial/showcase begins 2 minutes in.

The following regex works for any query parameters.
^https?:\/\/(www\.)?youtube(-nocookie)?\.com\/embed\/[a-zA-Z0-9_-]{11}(\?(&?\w+=(\w|\d)+)+)?$
(This setup then works as a default blacklist approach, where any non-wanted query parameters should be added to the remove section).
Simply due to the nature of regex, that requires the position of strings to match, or just matching 3 parameters would end up in a very clonky a,b,c | a,c,b | c,a,b | etc... (9 options total)

New changes in this PR are:
- Fixing the autoplay remover
- Modifying the regex to avoid a ReDOS

### DIFF
--- a/helpers/parse.js
+++ b/helpers/parse.js
@@ -44,7 +44,7 @@ export const configuredXss = new xss.FilterXSS({
                 // case where its ?followed=by&another=queryParam
                 if (index + remove.length < value.length && value.charAt(index + remove.length) === '&') {
                   value = value.replace(`${remove}&`, '');
-                } else {
+                } else if (index + remove.length >= value.length) {
                   value = value.replace(`?${remove}`, '');
                 }
               } else {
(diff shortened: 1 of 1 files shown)


falseresync (approved):

triphora (approved):

triphora:
Reverted this after some internal discussion.
If YouTube adds any other query params that we don't want to allow to be enabled by default, it would probably take us a while to add it to the blacklist. Especially if something can be abused, it is not something we want to have any chances of permitting. Instead, the whitelist should be expanded with any other params that may show up (like wmode), and this is best accomplished by using the URL API in JS.

PR #1082, opened 2023-04-06
Redesign a new visual theme switcher and store the preferred dark theme

### LINKED ISSUE #365: Change dark mode behaviour for theme toggles
Currently, the theme toggler in the navbar switches only between light and dark mode. There's also a hidden OLED mode accessible only from being logged in.

Instead of changing between all three (light, dark, OLED) options in the settings and only between the two (light, dark) in the navbar, the setting should let you choose whether your dark mode should be standard dark or OLED dark, so that the navbar toggler switches between light mode and the chosen dark mode.

Prospector:
Screenshot:
image

Closes MOD-358
Closes #365

### DIFF
--- a/assets/styles/components.scss
+++ b/assets/styles/components.scss
@@ -19,7 +19,7 @@
 
 .multiselect {
   color: var(--color-text) !important;
-  outline: 2px solid transparent;
+  outline: none !important;
 
   input {
     background: transparent;
@@ -38,6 +38,7 @@
     cursor: pointer;
     padding-left: 7px;
     padding-top: 10px;
+    outline: 2px solid transparent;
 
     transition: background-color 0.1s ease-in-out;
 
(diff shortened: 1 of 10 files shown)


triphora (changes requested):
Is there no way to reset it back to System in the UI after you've selected a theme? I know a few people who use dark mode at night and light mode during the day (not necessarily for Modrinth but for other applications in general), and just set a single browser/system toggle for it. I can see someone getting annoyed if they set something to one of the provided themes and then can't reset it back to system without clearing cookies

https://user-images.githubusercontent.com/60205699/232642943-67d8b453-28cc-4f8e-80ed-bc790e9b6f5f.mp4

Prospector:
yeah, when this PR was made system theme was removed because it wasn't working. it's been fixed since then so it'll be added back

Minenash:
Any update on this, this would be cool to see on the site

Prospector:
yeah, a better version of this will be implemented in The Redesign

PR #1085, opened 2023-04-07
Fix duplicate version featured links being wrong sometimes

### LINKED ISSUE #842: Adding multiple featured versions with the same version number for different platforms sometimes links to wrong page
### Describe the bug

The issue is a bit confusing so I'm not sure if the title is completely clear, but basically if you have two versions with the same version number (let's say 1.2.3) and the only difference between them is the platform supported (for example Fabric and Forge), it's possible for the featured version link for one to go to the wrong page, because the end of the URL will be `/version/1.2.3` instead of `/version/uniquestring`.

I'm just going to link to the discussion thread on Discord because it's more clear there with specific examples:
https://discord.com/channels/734077874708938864/783153806531100673/1061543051845185586

Prospector already has a proposed solution which is to make the featured version linking work the same way as it does in the versions tab (which works correctly in this case).

### Steps to reproduce

1. Go to https://modrinth.com/plugin/chunkyborder
2. Click on the featured version labelled "ChunkyBorder 1.1.34" for Fabric 1.19-1.19.2
3. you are redirected to the version for Bukkit, Spigot, and Paper instead

### Expected behavior

3. links to the version for Fabric, as it does in the versions tab

### Additional context

Apparently this only happens with "manually featured" versions. More context in the discord thread linked previously.

Prospector:
This may seem like a bit of an odd way to fix it but it works well. Left a note in the code so we know why it's done that way

Fixes #842
Closes MOD-357

### DIFF
--- a/pages/[type]/[id].vue
+++ b/pages/[type]/[id].vue
@@ -834,7 +834,14 @@ if (
 }
 
 versions.value = data.$computeVersions(versions.value, allMembers.value)
-featuredVersions.value = data.$computeVersions(featuredVersions.value, allMembers.value)
+
+// Q: Why do this instead of computing the versions of featuredVersions?
+// A: It will incorrectly generate the version slugs because it doesn't have the full context of
+//    all the versions. For example, if version 1.1.0 for Forge is featured but 1.1.0 for Fabric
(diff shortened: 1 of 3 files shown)


Geometrically (approved):

pop4959:
Thank you! ❤️

PR #1092, opened 2023-04-11
Add indicators for visibility on project settings page

mineblock11:
## Information

Replaces the following text:

> Set the visibility of your project. Listed and archived projects are visible in search. 
Unlisted projects are published, but not visible in search or on user profiles. 
Private projects are only accessible by members of the project.

With some indicators that are much easier to understand.

Potentially could word differently? Not sure if the following should be used instead for the search and member profile bulletpoint

- Visible in search results
- Visible on member profiles

### DIFF
--- a/pages/[type]/[id]/settings/index.vue
+++ b/pages/[type]/[id]/settings/index.vue
@@ -143,7 +143,7 @@
       </template>
       <div class="adjacent-input">
         <label for="project-visibility">
-          <span class="label__title">Visibility</span>
+          <span class="label__title">Target Visibility</span>
           <span class="label__description">
             Listed and archived projects are visible in search. Unlisted projects are published, but
             not visible in search or on user profiles. Private projects are only accessible by
--- a/pages/[type]/[id]/settings/index.vue
+++ b/pages/[type]/[id]/settings/index.vue
@@ -143,7 +143,7 @@
       </template>
       <div class="adjacent-input">
         <label for="project-visibility">
-          <span class="label__title">Target visibility</span>
+          <span class="label__title">Visibility</span>
           <span class="label__description">
             Listed and archived projects are visible in search. Unlisted projects are published, but
             not visible in search or on user profiles. Private projects are only accessible by
--- a/pages/[type]/[id]/settings/index.vue
+++ b/pages/[type]/[id]/settings/index.vue
@@ -145,7 +145,9 @@
         <label for="project-visibility">
           <span class="label__title">Visibility</span>
           <span class="label__description">
-            Set the visibility of your project.
+            Listed and archived projects are visible in search. Unlisted projects are published, but
+            not visible in search or on user profiles. Private projects are only accessible by
+            members of the project.
             <ul class="visibility-info">
               <li>
                 <TickIcon
--- a/pages/[type]/[id]/settings/index.vue
+++ b/pages/[type]/[id]/settings/index.vue
@@ -150,23 +150,23 @@
             members of the project.
             <ul class="visibility-info">
               <li>
-                <TickIcon
+                <CheckIcon
                   v-if="visibility === 'approved' || visibility === 'archived'"
                   class="good"
                 />
-                <CheckIcon v-else class="bad" />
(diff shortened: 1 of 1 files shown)


Prospector:
I don't like how this doesn't tell you what each does until you click on it. You would have to select Unlisted to know what it did, for example. That's not intuitive because you'd want to know what it did before you selected it

mineblock11:
Hm. Doesn't the user need to confirm their selection anyways via the "Save changes" button? Currently it's safe to switch between dropdown options without causing the visibility of the project to update.

mineblock11:
I forgot to upload a video of it as well:

Prospector:
Yeah but that's not intuitive to have to switch between options in order to figure out what they do. I think the display you added is fine, I'd just leave the old text (or an improvement upon it) as well.

mineblock11:
I've updated it with a better description.

pages/[type]/[id]/settings/index.vue
```diff
@@ -218,6 +253,9 @@ export default defineNuxtComponent({
     UploadIcon,
     SaveIcon,
     TrashIcon,
+    CrossIcon,
+    TickIcon,
```
Comment (Prospector) [inline pages/[type]/[id]/settings/index.vue:257]: I would rename this to CheckIcon to be more consistent but otherwise LGTM

Prospector (approved):

Prospector:
Wait, why is exit.svg even a thing? We use `x.svg` for everything it seems.

Minenash:
A problem with using a checklist looking system, besides what pro has mentioned, as that it could give the user a false impression of the actual state of the actual visibility (instead of the target visibility).

A possible idea would be to do something similar to the theme selector. Where each entry has information about it. Like it being listed in search/profile, who can see it, any warnings that the status displays

mineblock11:
The actual visibility of the project is always listed underneath its name isn't it? I assume with the description above the checklist it would make it obvious that it's the target visibility.

A fix could be renaming it from Visibility -> Target Visibility

Prospector (commented):
removing my approval

Prospector (changes requested):
why can't you just remove an approval

mineblock11:
Should be all good now

pages/[type]/[id]/settings/index.vue
```diff
@@ -143,11 +143,43 @@
       </template>
       <div class="adjacent-input">
         <label for="project-visibility">
-          <span class="label__title">Visibility</span>
+          <span class="label__title">Target Visibility</span>
```
Comment (triphora) [inline pages/[type]/[id]/settings/index.vue:146]:
```suggestion
          <span class="label__title">Target visibility</span>
```

triphora (commented):
Might not be worth fixing but there is a ton of unnecessary padding here
image

Prospector:
I don't like the heading "Target visibility". Maybe instead, when your selected visibility is different from the current saved visibility, it changes the text from "Visible in search" to "Will be visible in search" to show that it's unsaved.

Prospector (changes requested):
I don't like the heading "Target visibility". Maybe instead, when your selected visibility is different from the current saved visibility, it changes the text from "Visible in search" to "Will be visible in search" to show that it's unsaved.

Prospector:
I think that's pretty good!

Prospector:
Just rename back to Visibility

brawaru:
I personally don't like text changing like this, but also when thinking about solution I can only remember a couple UIs that show a table (current / new), others instead use a popup that explains consequences (e.g. GitHub), since changing visibility is a bit dangerous action.

Screenshot of table layout concept showing two cards below the description for visibility option, where first card shows current visibility (not visible in search, nor via profile, and can only be accessed by members through URL) and new visibility (visible in search and on profile, and accessible through URL by everyone. In URL visibility who can view URL is marked with bold text. On the right arrows and square is drown above tho selector, suggesting that it should be moved slightly above.

mineblock11:
This wouldn't fit on mobile, and I think its quite excessive for something so simple

Geometrically (approved):

PR #1102, opened 2023-04-18
Progressive Web App support

mineblock11:


mineblock11:
Going to open up pull requests, because taking a look at the OneSignal notification API - it's **quite** a bit of effort and would likely need you to rewrite a few parts of the existing notification system on labrinth - also, it wouldn't be free for modrinth 


https://onesignal.com/pricing

PR #1105, opened 2023-04-19
Extract Message banner component

falseresync:
Screenshot 2023-04-19 205457

I have extracted this part of the layout into its own component. Additionally I have fixed the pesky border rounding.


triphora:
The rounding was unintentional at first but I kinda think it's nice and unique

falseresync:
I agree that there’re better names for it, though “flash messages” is used often for something like that in web dev (usually for messages delivered via the session, see Rails or Symfony for example). Banner seems a bit too vague though. If you have any other suggestions or arguments I am open

As for the border, I stand by my opinion, and would like a third voice

Geometrically:
I like this PR's border better

pages/[type]/[id].vue
```diff
@@ -427,19 +427,15 @@
           :auth="auth"
           :tags="tags"
         />
-        <div v-else-if="project.status === 'withheld'" class="card warning" aria-label="Warning">
+        <MessageBanner v-if="project.status === 'withheld'" message-type="warning">
```
Comment (triphora) [inline pages/[type]/[id].vue:430]: Keep this one as an else-if, it shouldn't be displayed to members of the project

triphora:
Will be closed soon if comment isn't addressed and conflicts aren't resolved

falseresync:
ah, oi, I forgot, sorry; I was away from town. Going to fix soon!

triphora (approved):

PR #1106, opened 2023-04-20
Always use https URL when accessing wsrv.nl

Justsnoopy30:
This is necessary for a CSP PR on theseus that I'll open in a moment.


Geometrically (approved):

Geometrically:
closed since this has fixed by another PR

PR #1117, opened 2023-04-23
Renovate config files

falseresync:
ESLint and  Prettier config files used some old settings. Additionally I've fixed some linting issues so that I could remove a couple of ESLint rules.

Most of the changes are to Nuxt config. I extracted constant values into the front to avoid repetition and potential errors from misspelling identical properties. Plus I've migrated the file to TypeScript


.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
brawaru, line 1: Any particular reason to switch to JS config format? JSON currently has an advantage that editors can suggest the properties and check values if using Prettier extension. If using JS format it's probably worth adding the following comments to preserve suggestions, but still feels unnecessary, as it doesn't really change anything (?).

```suggestion
// @ts-check
/** @type {import("prettier").Options} */
module.exports = {
```

falseresync: For ESLint it's because in the future the JS format will be [updated and preferred](https://eslint.org/docs/latest/use/configure/configuration-files-new)

For Prettier there was no reason

For Nuxt - TypeScript is the new default, and additionally it'd be best to use TypeScript everywhere

falseresync: Actually, as I am now checking it, the new ESLint config can already be used without an env var

brawaru: Yeah other stuff is understandable and nice, that's why I asked under Prettier file change :ioa:

falseresync: I've reverted the Prettier file and added a type annotation for ESLint, just in case :)

triphora (approved):

triphora (changes requested):
The lint check in the CI appears to fail. Could that be fixed? Looks good otherwise

brawaru:
I noticed we have a leftover `jsconfig.json` btw, maybe we can get rid of it? I am not sure what the effects of it, but I am assuming `tsconfig.json` gets the priority, and it already covers the importing of JS files and aliases...

falseresync:
> I noticed we have a leftover `jsconfig.json` btw, maybe we can get rid of it? I am not sure what the effects of it, but I am assuming `tsconfig.json` gets the priority, and it already covers the importing of JS files and aliases...

Oh yeah, this is true https://code.visualstudio.com/docs/languages/jsconfig

brawaru:
No way adding comment just worked

triphora:
@brawaru

```diff
- <script setup lang="ts">
+ <script setup>
```

triphora (approved):

app.vue
```diff
@@ -5,7 +5,8 @@
     <NuxtPage />
   </NuxtLayout>
 </template>
-<script setup lang="ts">
+<script setup>
+// For some reason, this file cannot be TypeScript. Why?
 import ModrinthLoadingIndicator from '~/components/ui/modrinth-loading-indicator'
 import Notifications from '~/components/ui/Notifications'
```
brawaru, line 11:
```suggestion
<script setup lang="ts">
import ModrinthLoadingIndicator from '~/components/ui/modrinth-loading-indicator'
import Notifications from '~/components/ui/Notifications.vue'
```

You're supposed to add `.vue` extensions to SFC imports.

Geometrically (approved):

PR #1160, opened 2023-05-21
Support for loader + game version combos

### LINKED ISSUE #447: Filtering loaders and game versions together does not specifically look for them on the same version
### Environment

Website

### Describe the bug

If you select a specific mod loader and version (for example "forge" & "1.7.10") you get all mods that have a forge version and a 1.7.10 version. This includes mods whose 1.7.10 version isn't on forge.

For Example:
image

Of the mods listed here the odd one out is Alternate Current, which has a 1.7.10 Legacy Fabric version but not a forge version (the only forge version of the mod listed is for 1.18)

### Steps To Reproduce

1 Go to https://modrinth.com/mods
2 Select "Forge" in Loaders
3 Select "1.7.10" in versions
4 Click on "Alternate Current"
5 Click on "Versions"
6 Select "Forge" as Loader
7 Only 1.18 is shown

### Expected behavior

Only mods whose 1.7.10 version uses Forge should show up

### Additional context

_No response_

stairman06:
Resolves #447, requires https://github.com/modrinth/labrinth/pull/622 on the backend.

This modifies the search query and uses a new query parameter `a` for loaders. The previous `g` for OR facets works for backwards-compatibility, but does not show up in the UI.

### DIFF
--- a/pages/search/[searchProjectType].vue
+++ b/pages/search/[searchProjectType].vue
@@ -770,6 +770,32 @@ export default defineNuxtComponent({
       if (index !== -1) {
         this.selectedLoaders.splice(index, 1)
       } else {
+        if (elementName === 'purpur') {
+          if (!this.selectedLoaders.includes('paper')) {
+            this.selectedLoaders.push('paper')
+          }
+          if (!this.selectedLoaders.includes('spigot')) {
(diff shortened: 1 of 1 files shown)


triphora:
Fixed as of modrinth/labrinth#731

PR #1161, opened 2023-05-23
Support more spec in mc version range parsing

ThatGravyBoat:
### Description
This changes the semver parsing for Fabric and Quilt infer to support more of the spec properly, while this doesn't support the whole spec it does support a large majority of it that is used in regards to Minecraft.

### Issues 
This does not support all of the spec but should allow expansion later, it currently does not support the following:
- Hyphen Ranges
- Comparator Sets
- Disjunctions
- X Ranges in tilde or caret ranges

All of which wasn't supported before or weren't done correctly.

### Downfall

This should also reduce the amount of mishaps as it is stricter with how it operates so for example it won't allow `>=1.19.3-` to parse to `>=1.19` as it requires it to match the spec more accurately. While this may mean that some users using more advanced version ranges will need to put in their versions manually it will likely be faster or the same amount of time as it won't add unintended versions to the versions list.


Prospector:
Could you run the fix script, seems to be failing linting

ThatGravyBoat:
The other PR that superseded this was merged.

PR #1168, opened 2023-05-28
Switch to pnpm

### LINKED ISSUE #1135: Reconsider the decision to use yarn in favour of modern alternatives like pnpm

Not so long ago Modrinth repositories were switched to use legacy `yarn` v1 while previously using `npm`. I think this decision was shortsighted, misinformed and suboptimal.

One of the huge problems with legacy Yarn is that it's now frozen and not actively maintained, it **has not been maintained since 2021**-2022 and they recommend to migrate to Berry (v2+). And it already broke on my pull request that does not do anything unusual, which can be confirmed by simply using npm and seeing that it works with no issue, and I don't even know how to mitigate this!

Another problem I have encountered is Yarn's terrible caching strategy. Back in the days Yarn was much more space efficient than npm, but this has since changed. During active usage of npm my its cache store only accumulated 2 GB, whereas solely by working with Knossos my Yarn cache directory accumulated 4.3 GB.

The culprit? Yarn ignores the system requirements for packages and installs them even though they're clearly marked as incompatible with current OS. Therefore you end up with binaries of esbuild for all operating systems (including freaking Android!) for several versions of esbuild. 

Another culprit is that Yarn doesn't cache individual files but rather whole packages. This is very space inefficient since you end up with lots of duplications of multiple versions of the package that haven't changed much between these versions.

Other issues I had with Yarn is it getting completely stuck on certain steps and not showing anything even with verbose logs, but even I find the reason, this will never get fixed anymore, as the yarn is frozen (effectively archived).

I don't think Yarn Berry is the right choice for Modrinth repos either as we rely one some of npm's quirks such as dependency hoisting. I'm saying ‘we’, but not like its our choice, it's a [problem with Nuxt](https://github.com/nuxt/nuxt/issues/14146), really.

In my short testing Yarn Berry has failed to work with Knossos because it expects strict dependencies, another thing that Nuxt requires to be set to disabled, but I haven't found option to do this with Yarn Berry.

So given we want to preserve some of the npm's quirks, but don't want to use npm due to how slow and space inefficient it is, I would propose switching to pnpm, which is gaining popularity among JavaScript developers.

> pnpm is a fast and disk space-efficient package manager that serves as a drop-in replacement for npm. It centralizes library code storage, enabling shared use of dependencies across multiple projects, which leads to better performance and reduced disk space usage. pnpm installs packages into a single `.pnpm-store` subdirectory in your home directory and uses symlinks to link packages into the projects where they are needed. This also results in faster project initialization times compared to npm, even if npm has cached the package. [flaviocopes.com](https://flaviocopes.com/pnpm/), [blog.logrocket.com](https://blog.logrocket.com/managing-full-stack-monorepo-pnpm/)
>
> **Note** Answers by AI may not always be accurate.

Previously Geo dismissed pnpm calling it annoying, but I'm yet to see the reasons why it's annoying. Perhaps it was misconfigured and caused trouble with Nuxt, but there's a `.npmrc` config now which includes stuff like `shamefully-hoist`, which will be respected as long as pnpm is ran from the root of the project.

pnpm also has great DX shortcuts like Yarn:
- you can run `pnpm [script / bin]` to quickly run commands like `dev` (`pnpm dev`) or binaries from project's dependencies like `nuxi` (`pnpm nuxi prepare`)
- it has nice aliases like `pnpm i` (`pnpm install`) or `pnpm b` (`pnpm run-script build`), and also `pnpm add [pkg]` (`pnpm install [pkg]`)
- there are utilities to quickly update packages like `pnpm update -i`, or to see why certain dependency is present `pnpm why [dep]`
- it allows to quickly patch broken packages using `pnpm patch [dep]`.

My personal experience with pnpm is incredible and I can not recommend it to people enough!

> **Note** This issue is not sponsored by pnpm 😄

**Switching back to npm**
npm is very inefficient and slow. It received a lot of improvements and very close to legacy Yarn compared what it was, but I still think it's far far from great. However, it's still a better option than legacy Yarn, because it's continuously maintained, so if pnpm does not work, it can be an option.

**Solving the problem and migrating to Yarn Berry**
Yarn currently promotes completely different model of Plug n' Play dependencies that aren't widely supported (and especially not supported by Nuxt), so it has to be configured to use npm linker, which is okay, but then there's the issue with peer dependencies. Yarn Berry is also pretty different from Yarn.

I honestly don't think it's worth the hassle to get it working, especially with pnpm and npm working right now without additional configuration (asides from what we have).

### Additional context

I'm publishing this in Knossos repo because Omorphia and Theseus are out of our concern right now, and such switch might be disruptive to productivity there anyway, whereas the progress here for now is pretty chill allowing for experimentation like this.

Thanks for coming to my TED talk!

brawaru:
## Summary

This pull request switches the package manager used to work with the repo to pnpm.

## Why is this change necessary?

Yarn v1 (classic) has not been maintained since 2021 and has severe issues when working with modern codebases such as Knossos. It fails on [the basic sets of dependencies](https://github.com/modrinth/knossos/actions/runs/4962424915/jobs/8880460564), [it ignores basic package installation constraints and bloats the cache directory](https://github.com/modrinth/knossos/assets/10401817/a74d6db0-e293-4648-8b96-82d74a521dd9), it's slow and inefficient compared to alternatives.

## Why pnpm?

pnpm has been around for quite some time and gained huge popularity in recent years. It is praised for its space efficiency, speed, and great developer experience. It also uses isolate (and more correct) model for installing packages, however [this does not work with Nuxt](https://github.com/nuxt/nuxt/issues/14146), so it's not applicable to our use case.

[Learn more about pnpm on its official site →](https://pnpm.io/)

## What changed?

- Yarn lock file has been replaced with **fresh** pnpm lock file. This means some dependencies will be updated by this pull request.
- pnpm will be specified in `package.json` as package manager required to work with Knossos. It has to be updated periodically. This allows to use [`corepack`](https://nodejs.org/api/corepack.html).
- CI uses `corepack` to install pnpm and runs scripts through pnpm.

## Additional information

The repository already has correctly set up `.npmrc` to enable shameful hoisting of dependencies.

If there are overrides for commands in Vercel configuration, they should be changed to use pnpm.

Resolves #1135


mineblock11:
Does the CI cache the pnpm dependencies between runs?

brawaru:
@mineblock11 of course:

Screenshots of three different caches: first one is for pnpm (ran for this pull request), which has length of just 57 MB, the second and third are for yarn (one for pull request 738, the other for master branch)

Geometrically (approved):

PR #1177, opened 2023-06-05
Draft: Category descriptions

bluebear94:
See #1166.

Currently, I’ve added a function to get the description for a category, but I still need to figure out how to show them in the UI.


Minenash:
One idea for visuals that doesn't add 20+ sentences, would be to have an area that gives the description of the currently hovered or focused tag

Prospector:
This would be a good idea to do in the future possibly with tooltips or something. Now that we have translations, we'd want these to be translatable for sure. Doesn't seem like this particular PR needs to be open anymore.

PR #1189, opened 2023-06-11
Fix CTRL clicking on versions issue

Flowko:
Fixes the issue Described here: #1184 

Fix:
- Converted the `@click` and the div element to a `NuxtLink` that will act as an ancher.


triphora (changes requested):
Please run `pnpm run fix`

Flowko:
@triphora done

triphora (approved):

PR #1191, opened 2023-06-11
fix: featured version ui shifts

Flowko:
### Issue:
after the change i did in #1189 it broke the layout of the featured version.
Screenshot 2023-06-12 at 12 09 57 AM

the final DOM made no sense, as the content inside of the main `NuxtLink` was being generated outside of it, the fix was to only render it on the client side, as i believe it was a server render mismatch issue.

Fixed version:
Screenshot 2023-06-12 at 12 11 54 AM


triphora (approved):
I love Chromium-only issues. (/s) This was not an issue when I tested the other PR on Firefox but now I see the issue on Chromium.

PR #1200, opened 2023-06-13
Use server routes to supply client data

brawaru:
Re-building and re-deploying the site every thirty minutes just to update data is very expensive and hurts user experience by constantly blowing out the client's cache without meaningful changes to the bundle. Even ignoring that, hardcoding changing data, even though it would be rebuilt eventually is considered a very bad practice. Overall, this all feels like a terrible hack.

This pull request should address this issue by removing client data generation logic and instead creating the supply logic using Nitro server cached routes that will refresh on the server every thirty minutes unless there's a re-deploy due to actual changes.


brawaru:
Redrafting to add support for persistent storage (KV) for workers and resolve conflicts.

brawaru:
It seems that this approach is not compatible with short-living workers or serverless functions:

- You can, of course, use [Workers KV](https://developers.cloudflare.com/workers/learning/how-kv-works/), but reads on it will probably be quite expensive.
- Also, when using serverless functions insead of workers, you cannot use KV bindings directly and have to rely on KV over HTTP, which doesn't have low latency benefit. There also seems to be a bug in Unstorage with KV over HTTP where it simply doesn't read nor store anything.
- Using Redis could've been [possible](https://unstorage.unjs.io/drivers/redis), but establishing connections may get also expensive and suboptimal.

So it seems there is no good solution to this problem, expect for directly contacting API to fetch tags and homepage projects, but this may be taxing on the API if no proper caching is implemented for these endpoints, there is also no `/homepage` API endpoint at this time.

Closing this pull request. Hopefully some better solution comes up in the future to replace the re-deployment hack.

brawaru:
Reopening as according to Geo KV won't cost much. I have implemented KV support for Cloudflare and Vercel, below are steps to set it up properly. If cache is not set up then it falls back to in-memory storage that is not persistent in workers. Additional persistent storage drivers should be relatively easy to add, check out [Unstorage documentation](https://unstorage.unjs.io/) for the list of all storage options and custom storage driver implementation steps.

Additionally also implemented a cached overlay driver that will store KV response in memory for the lifespan of a worker to prevent excessive and stale KV reads. Writes aren't immediately flushed with Cloudflare KV for whatever reason and may return stale data making Nitro re-compute the response and write it back to KV (effectively making caching ineffective), so having a cached overlay somewhat protects against this. #WondersOfModernWebDevelopment

<details>
<summary>Setting up Cloudflare KV (Pages)</summary>

1. Open Workers & Pages Cloudflare Dashboard.
2. Switch to KV page.
3. Create a namespace, call it anyhow you like.
4. Go back to Overview, and open Knossos application.
5. Go to Settings → Functions, scroll until ‘KV namespace bindings’.
6. Click Edit bindings, Add variable, call it whatever (e.g. CACHE), and set it to KV namespace you've created. Save.
7. Now go to Environment variables. Edit variables for production, add below and save.
  - `CACHE_STORAGE_OPTION` = `cloudflare-kv`
  - `CF_KV_BINDING_CACHE` = \<name of the binding variable\>
8. Retry latest deploy (if this PR was merged before setting up KV).

> **Note**
> To check that cache is functioning you can go to KV page again, click on the namespace, and check that keys like `cache:nitro:handlers:homepage.<hash>.json`, `cache:nitro:handlers:tags.<hash>.json` are present.

</details>

---

<details>
<summary>Setting up Cloudflare KV (Workers)</summary>

1. Open Workers & Pages Cloudflare Dashboard.
2. Switch to KV page.
3. Create a namespace, call it anyhow you like.
4. Copy ID of your namespace.
5. Edit `wrangler.toml` to include:
   ```toml
	[[kv_namespaces]]
	binding = "<name of binding variable>"
	id = "<ID of the KV namespace>"
	```
6. Edit deployment workflow to build Nuxt with environment variables:
   - `CACHE_STORAGE_OPTION` = `cloudflare-kv`
   - `CF_KV_BINDING_CACHE` = \<name of the binding variable\>
7. Push your changes and await for re-deployment.

> **Note**
> To check that cache is functioning you can go to KV page again, click on the namespace, and check that keys like `cache:nitro:handlers:homepage.<hash>.json`, `cache:nitro:handlers:tags.<hash>.json` are present.

</details>

---

<details>
<summary>Setting up Vercel KV</summary>

1. Open Knossos project page.
2. Switch to Storage tab.
3. Press Connect Store / Create New Database.
4. Choose KV (Durable Redis).
5. Configure it how you want (e.g. by adding read replicas in other regions).
6. On the next screen (when asked about environments and environment variables prefix) leave everything as is. Press Connect.
7. Switch to Settings tab, go to Environment variables.
   1. Ensure that the following variables are now present:
      - `KV_URL`
      - `KV_REST_API_URL`
      - `KV_REST_API_TOKEN`
      - `KV_REST_API_READ_ONLY_TOKEN`
   2. Add variable `CACHE_STORAGE_OPTION` with value `vercel-kv`.
8. Redeploy latest deployment (if this PR was merged before setting up KV).

> **Note**
> To check that cache is functioning you can go to Storage tab again, click on the KV database, and type `KEYS cache*` in the CLI. It should print something like `cache:nitro:handlers:homepage.<hash>.json`, `cache:nitro:handlers:tags.<hash>.json`.

</details>

---

Here's an example app running a similar configuration of KV:
https://github.com/brawaru/nuxt-cf-test-app

Live proof of work:

- Cloudflare Pages
  - https://nuxt-cf-test-app.pages.dev/api/timestamp (10s)
  - https://nuxt-cf-test-app.pages.dev/api/long-timestamp (24h)
- Cloudflare Workers
  - https://nuxt-cf-test-main.alexander-sorokin.workers.dev/api/timestamp (10s)
  - https://nuxt-cf-test-main.alexander-sorokin.workers.dev/api/long-timestamp (24h)
- Vercel
  - https://nuxt-cf-test-app.vercel.app/api/timestamp (10s)
  - https://nuxt-cf-test-app.vercel.app/api/long-timestamp (24h)

brawaru:
Saw merge conflicts, tried to resolve them, but figured why not take the opportunity and decided to re-create this almost from scratch to make it much cleaner and maintainable 👍🏻  

Pretty much ready for review

plugins/shared-state.ts
```diff
@@ -0,0 +1,27 @@
...
+    provide: {
+      ...provideSharedStates({
+        static: {
+          tags: await useTagsFetch(),
+        },
+        dynamic: {
+          homepageProjects: defineDynamicStateHolder({
+            key: 'homepage-projects',
+            state: useFetch('/api/homepage-projects'),
+            ttl: 10 * 60,
```
falseresync, line 15: TTL'ing it on the server route would be enough

brawaru: Not really, no, this is client's own TTL, which may not match the TTL on the server.

falseresync: It may not, but this scenario is not nearly complicated enough to grant a need for 2 separate TTLs

composables/shared-state.ts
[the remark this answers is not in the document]
falseresync: Nuxt Docs on useFetch explicitly says it doesn’t do duplicate requests on hydration

Bringing Pinia in would be a good idea anyway, if this really can’t be done without a store. Other parts of the application will benefit from it down the line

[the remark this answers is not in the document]
brawaru: Docs can say whatever, we tested, it does refetch on hydration and every time composable is used. It's stupid and I think there's a bug report for that somewhere.

plugins/shared-state.ts
brawaru: On the server this TTL is never used, it's purely clint-side, server has TTL for the route for 30 minutes after the first invocation. Here the client will consider data stale after 10 minutes and try to fetch again. I mean it can be changed to match the server, but again, there will be mismatch between when server first cached the route and when client got that data, but not that homepage projects is such a critical data to re-fetch frequently.

composables/shared-state.ts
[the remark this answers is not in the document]
falseresync: And that separate PR will argue the exact same thing, that it’s niche and therefore needs to implement it all separately again. Given that Modrinth is an understaffed start-up, developer’s time matters more than bundle size

plugins/shared-state.ts
falseresync: You misunderstood. There’s no much need to store it client-side with a TTL at all, the payload size is negligible and the amount of visits to the homepage is small

composables/shared-state.ts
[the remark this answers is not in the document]
brawaru: It's not about bundle size, it's about ability to review, it takes time to review stuff like this, plus Modrinth loves to overhaul lots of files in a single feature branch and then merge it all at once, and then you get bunch of conflicts and have to rebase all your changes, sometimes redoing them from scratch, because some things fundamentally changed. So the smaller the PR the better, for them, for you. If you make a PR focusing on adding Pinia and it gets merged, then you can make another one migrate one thing, then another, and another, I don't think this will be an issue.

triphora:
We talked about this internally, and it's probably just better to convert it back to the old way of requesting everything on the client - ever since redis caching was added to the backend, it wouldn't affect performance any more or less than this would, from our understanding.

brawaru:
Addressed the concerns both of you had, @falseresync and @triphora. Reapplied simple strategy and just removed all of the build-time caching as you asked. There's no local caching or anything like that. Now the server will make a request and serve it to client as a payload, the client will refetch it every time homepage or Modrinth app page are open.

Since then I learnt a bit about ISR (Incremental Static Regeneration) which we could've used for the API routes (so they just regenerate every x minutes and in a meanwhile serve the same data to everyone), but Cloudflare Pages don't support this technology, so it essentially would end up what I made in the first iteration — caching with KV.

falseresync (approved):

PR #1202, opened 2023-06-14
Add sort by dropdown in user projects

kevoh1516:
Added a "sort by" dropdown on the user project page.


Prospector (changes requested):
Not a fan of the placement of this, it should be above the list of versions somewhere. Don't have time to dedicate to this rn though

PR #1210, opened 2023-06-17
Add the language setting page

brawaru:
Adds a settings page that allows to change the language. Includes all the fancy stuff like search, categories, shimmering for when the language is loading, and scary red message for when it fails to load.

Screenshot of the page with Pirate language enabled

Screenshot of language loading error


Q2297045667:
This PR is great

Mysterious-Dev:
Any news for this PR ?

Geometrically (approved):

Prospector:
this shouldn't be under User settings, and IMO the page should be entirely hidden for now because we don't have localization anywhere.

image

PR #1225, opened 2023-07-01
RSS feeds, continued

BasiqueEvangelist:
mineblock11 seems to have stopped working on #1159. This PR is a continuation of that PR, with some additional changes.

- [x] User notification feed removed (for now)
  Just passing the token via a header feels like it wouldn't be implemented by many people, and putting the GitHub token in the path seems a bit worrying, especially if the URL gets leaked. Having a separate feed secret would be nicer, but that requires changes on labrinth.
- [x] Much more metadata is put in the feed
  The version items have their loaders and game versions as categories and the publisher is listed as the author, the project's team is listed in the feed, et cetera.
- [ ] Project versions page links to feed
  The most basic form of this is to link to them via `<link>` elements, but showing off the feed visually would also be pretty nice.


BasiqueEvangelist:
Ready for review, just in case modrinth/labrinth#527 isn't accepted

server/routes/feed/[feed_type]/project/[id].js
```diff
@@ -0,0 +1,101 @@
...
+      (await $fetch(API_URL + 'user/' + version.author_id))
+
+    const loadersPretty = version.loaders.map(capitalizeString).join(', ')
+    const gameVersionsPretty = version.game_versions.join(', ')
+
+    feed.addItem({
+      title: `${version.name}`,
+      id: `${projectUrl}/version/${version.id}`,
+      link: `${projectUrl}/version/${version.id}`,
+      description: `Loaders: ${loadersPretty}, Minecraft versions: ${gameVersionsPretty}`,
```
Comment (Prospector) [inline server/routes/feed/[feed_type]/project/[id].js:70]: if loadersPretty is comma-separated, shouldn't the separator between Loaders and MC versions be a semicolon? Tbh, not entirely sure this is the best description for these, but I am not entirely sure what is.

PR #1232, opened 2023-07-10
Threads and more!

Prospector:
More including:
- Overhauled moderation
- 2 way mod messages
- Resubmission confirmation modal
- Report threads
- New notification system

probably more


plugins/shorthands.js
```diff
@@ -1,3 +1,5 @@
+import { getProjectTypeForUrlShorthand } from '~/helpers/projects'
```
Geometrically, line 1: need to add .js ext here, see GH actions build

pages/moderation/messages.vue
```diff
@@ -0,0 +1,68 @@
...
+  app.$notify({
+    group: 'main',
+    title: 'Error loading inbox',
+    text: error.value,
+    type: 'error',
+  })
+  console.error(err)
+}
+
+function getLink(thread) {
```
Geometrically, line 60: unused var

pages/[type]/[id]/moderation.vue
```diff
@@ -0,0 +1,216 @@
...
+        <WarningIcon /> Messages are currently not available on this project.
+        <span v-if="config.public.apiBaseUrl.startsWith('https://staging-api.modrinth.com')">
+          This is expected on staging because the database has not been fully migrated, please make
+          a new project to try message threads.
+        </span>
+        <span v-else>
+          If you're seeing this, please report this as an issue including your project's ID:
+          <CopyCode :text="project.id" />
+        </span>
+      </div>
```
Comment (Geometrically) [inline pages/[type]/[id]/moderation.vue:84]: you can remove this, as threads are mandatory for reports + projects now

pages/[type]/[id].vue
```diff
@@ -1070,6 +1133,23 @@ function openModerationModal(status) {
...
+        members.forEach((it, index) => {
+          members[index].avatar_url = it.user.avatar_url
+          members[index].name = it.user.username
+        })
+
+        return members
+      },
+    }
+  )
+}
```
Comment (Geometrically) [inline pages/[type]/[id].vue:1151]: why is this needed? when would you update members outside of the members management page?

pages/[type]/[id].vue
```diff
@@ -689,6 +717,38 @@
...
+              :href="`https://staging.modrinth.com/${project.project_type}/${
+                project.slug ? project.slug : project.id
+              }`"
+              rel="noopener nofollow"
+              target="_blank"
+            >
+              <ExternalIcon aria-hidden="true" />
+              View on staging.modrinth.com
+            </a>
+          </div>
```
Comment (Geometrically) [inline pages/[type]/[id].vue:751]: imo this should be removed, I don't see a good use case for it, and it's simple to go to the URL if you really want to

components/ui/ProjectPublishingChecklist.vue
```diff
@@ -1,4 +1,18 @@
...
+    </p>
+    <div class="input-group">
+      <button class="iconified-button brand-button" @click="acceptInvite()">
+        <CheckIcon />Accept
+      </button>
+      <button class="iconified-button danger-button" @click="declineInvite()">
+        <CrossIcon />Decline
+      </button>
+    </div>
+  </div>
```
Geometrically, line 15: why is this in the publishing checklist? why not just in the main mod page?

components/ui/LoadingComponent.vue
Geometrically, line 1: what is this component used for? we already have a loading component for pages

components/ui/ConditionalNuxtLink.vue
```diff
@@ -0,0 +1,21 @@
...
+  to: {
+    type: String,
+    required: true,
+  },
+  isLink: {
+    type: Boolean,
+    required: true,
+  },
+})
+</script>
```
Geometrically, line 21: I don't think this needs to be a component. You can just use `<component :is="component" />

Geometrically (changes requested):

Geometrically (changes requested):
you also need to run prettier/lint

pages/[type]/[id]/moderation.vue
Comment (Prospector) [reply pages/[type]/[id]/moderation.vue:84]: seems better to have an error message than not, no? I could just simplify it though and remove the first part

pages/[type]/[id].vue
Comment (Prospector) [reply pages/[type]/[id].vue:751]: I use it literally all the time when working to compare pages to prod, don't see any reason to remove it since it doesn't show on prod or staging deploys?

components/ui/ProjectPublishingChecklist.vue
Prospector: I could rename the component, it's more of a project page header I suppose? It should show in both settings and main project pages, if it was just in the page it would be duplicated code. Or I could make it its own component? but don't really see any reason to separate it

components/ui/LoadingComponent.vue
Prospector: are you talking about the loading bar at the top of the screen? This is for when just a part of the page is loading, it makes navigation feel really smooth and instantaneous. Check out the new pages and feel how smooth navigating to them feels.

components/ui/ConditionalNuxtLink.vue
Prospector: think you forgot to finish writing this comment, not sure what you meant?

pages/[type]/[id].vue
Comment (Prospector) [reply pages/[type]/[id].vue:1151]: this is called when members accept invites

components/ui/ConditionalNuxtLink.vue
Geometrically: component :is (lookup vue dynamic comments)

components/ui/ProjectPublishingChecklist.vue
Geometrically: yeah renaming would probs be good

pages/[type]/[id]/moderation.vue
Comment (Geometrically) [reply pages/[type]/[id]/moderation.vue:84]: the message would never show up, because projects without threads are invalid in the DB and won't be returned. so the code can be removed

Geometrically (approved):

PR #1234, opened 2023-07-14
New authentication

Geometrically:
Fixes MOD-501


pages/auth.vue
```diff
@@ -0,0 +1,157 @@
...
+.auth-container {
+  width: 24rem;
+  padding: var(--gap-xl);
+  background-color: var(--color-raised-bg);
+  border-radius: var(--radius-lg);
+  margin: 2rem auto;
+
+  h1 {
+    font-size: 1.75rem;
+    font-weight: 500;
```
Prospector, line 21: the thin font looks ugly here imo, I'd just remove this and it'll use the standard heading weight

pages/auth.vue
```diff
@@ -0,0 +1,157 @@
+<template>
+  <div>
+    <NuxtPage class="auth-container card" :route="route" />
```
Prospector, line 3: since there's style conflicts with omorphia, use `universal-card` for now. something we'll have to deal with but probably out of scope for now. this will fix the padding on the card

pages/settings/account.vue
```diff
@@ -9,54 +9,259 @@
...
+              <span class="label__title">Verify code</span>
+              <span class="label__description"
+                >Enter the one-time code from authenticator to verify access.
+              </span>
+            </label>
+            <input
+              id="verify-code"
+              v-model="twoFactorCode"
+              maxlength="6"
+              type="number"
```
Prospector, line 182: 2FA backup codes aren't numbers (need to update in all the places)

pages/settings/account.vue
```diff
@@ -66,53 +271,83 @@
...
+        <div>
+          <button class="iconified-button" @click="showTwoFactorModal">
+            <template v-if="auth.user.has_totp"> <TrashIcon /> Remove 2FA </template>
+            <template v-else> <PlusIcon /> Setup 2FA </template>
+          </button>
+        </div>
+      </div>
+      <div class="adjacent-input">
+        <label for="theme-selector">
+          <span class="label__title">Manage authentication providers</span>
```
Prospector, line 340: Button already says manage, I'd just have the title here be Authentication providers

Prospector (changes requested):
"account information" in this case is just an email?
image

Table look really bad in light mode:
image

PR #1239, opened 2023-07-22
Fix bug with wsrv

triphora:


helpers/parse.js
```diff
@@ -67,6 +67,7 @@ export const configuredXss = new xss.FilterXSS({
     if (tag === 'img' && name === 'src' && !value.startsWith('data:')) {
       try {
         const url = new URL(value)
+        url.searchParams.delete('errorredirect')
```
brawaru, line 70: This would remove `errorredirect` from any link, which probably wouldn't hurt anyone (doubt anyone uses such parameter for any other link), but still a bad practice in general, and may confuse people looking at this code in the future. I think it's worth explicitly checking that the link's hostname is `wsrv.nl` or `www.wsrv.nl`.

```suggestion
        if (url.hostname === 'wsrv.nl' || url.hostname === 'www.wsrv.nl') {
          url.searchParams.delete('errorredirect')
        }
```
*(the suggestion is made outside the code editor, and not formatted with Prettier)*

Geometrically (approved):

PR #1248, opened 2023-07-29
Change version infering to use full semver range spec

ThatGravyBoat:
### What does this PR change?
This changes the infer to support all of semver by instead of handling semver ranges manually it uses the npm `node-semver` package.
This also adds support for forges maven ranges by converting maven ranges to semver ranges and passing them to the semver version range handling to get the game versions.

Supersedes #1161

### DIFF
--- a/helpers/infer.js
+++ b/helpers/infer.js
@@ -1,6 +1,7 @@
 import TOML from '@ltd/j-toml'
 import JSZip from 'jszip'
 import yaml from 'js-yaml'
+import { satisfies } from 'semver'
 
 export const inferVersionInfo = async function (rawFile, project, gameVersions) {
   function versionType(number) {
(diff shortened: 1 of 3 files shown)


helpers/infer.js
```diff
@@ -114,7 +164,7 @@ export const inferVersionInfo = async function (rawFile, project, gameVersions)
         loaders: ['fabric'],
         version_type: versionType(metadata.version),
         game_versions: metadata.depends
-          ? gameVersionRange(metadata.depends.minecraft, gameVersions)
+          ? getGameVersionsMatchingSemverRange(metadata.depends.minecraft, simplifiedGameVersions)
```
triphora, line 167: Looks good for the most part, but you forgot to apply this to quilt.mod.jsons too

triphora (changes requested):

helpers/infer.js
ThatGravyBoat: the second line of the `getGameVersionsMatchingSemverRange` checks if the input is a single range or an array of ranges and does it accordingly

triphora:
Just need to run `pnpm run fix` then it should be good

triphora (approved):

Geometrically (approved):

PR #1260, opened 2023-08-05
Update legal docs

Geometrically:


triphora (approved):
would be nice if we could split this up into separate sections in the nav in the future

CodexAdrian (approved):
LGTM

PR #1288, opened 2023-08-08
Fix bugs with threads

Geometrically:


helpers/notifications.js
```diff
@@ -134,10 +134,11 @@ export function groupNotifications(notifications, includeRead = false) {
   for (const notification of notifications.filter((notif) => includeRead || !notif.read)) {
     // Group notifications of the same thread or project id
     if (notification.body) {
+      console.log(notification.body)
```
triphora, line 137: console log

layouts/default.vue
```diff
@@ -476,6 +476,10 @@ export default defineNuxtComponent({
     },
   },
   mounted() {
+    if (process.client && window) {
```
triphora, line 479: are you sure it's safe to have client-side-only stuff like this?

pages/[type]/[id].vue
```diff
@@ -373,22 +373,6 @@
...
-          <a href="https://docs.modrinth.com/docs/modpacks/playing_modpacks/" :target="$external()"
-            >our documentation</a
-          >
-          which provides instructions on using
-          <a href="https://atlauncher.com/about" :target="$external()" rel="noopener"> ATLauncher</a
-          >, <a href="https://multimc.org/" :target="$external()" rel="noopener">MultiMC</a>, and
-          <a href="https://prismlauncher.org" :target="$external()" rel="noopener">
-            Prism Launcher</a
-          >.
-        </div>
```
Comment (triphora) [inline pages/[type]/[id].vue:391]: this definitely isn't worth removing entirely imo - there's really no other way on the site to tell people *how* to install a modpack

triphora (changes requested):

pages/[type]/[id].vue
Comment (CodexAdrian) [reply pages/[type]/[id].vue:391]: Agreed, perhaps replace it with text directing the user to install the new Modrinth App. Modpacks are still in beta support, and people might not be aware theres an official first party solution now

CodexAdrian (changes requested):
Emma on point here

pages/[type]/[id].vue
Comment (Prospector) [reply pages/[type]/[id].vue:391]: yup, I'd have it say to [download the Modrinth App] to install modpacks & link to the documentation for how to download them with other launchers

layouts/default.vue
Geometrically: yes its only client side code, it affects scroll behaviour

triphora (approved):

CodexAdrian (approved):

PR #1295, opened 2023-08-10
Replace project icon border

naqaden:
https://github.com/modrinth/knossos/issues/1280

Project pages with featured images and gallery-view search results have borders around their 96x96 icons, which squish them to 88x92. These box-shadow replacements preserve icon size while staying visually equivalent (rounded 4px border around left, top, and right).

edit: I've made a mess of this history because I'm limited to the browser and exploring options.


naqaden:
Too messy, will open a clean PR.

PR #1304, opened 2023-08-14
Add some text for when you haven't filed any reports

qtchaos:
Add some text for when you haven't filed any reports, helps the reports page fit in with the other dashboard pages.
image

Closes https://github.com/modrinth/knossos/issues/1270

### DIFF
--- a/components/ui/report/ReportsList.vue
+++ b/components/ui/report/ReportsList.vue
@@ -14,7 +14,7 @@
     :auth="auth"
     class="universal-card recessed"
   />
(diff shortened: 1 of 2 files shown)


Prospector:
Technically, I think the text should say something like "You don't have any active reports." because I believe it doesn't currently show historic reports, so the user could've reported something in the past but there's nothing there right now.

qtchaos:
image

Prospector (approved):

PR #1311, opened 2023-08-17
Improve download button in changelog

Nitrrine:
Before_And_After

I also fixed that `.markdown-body` style not applying normally

### DIFF
--- a/pages/[type]/[id]/changelog.vue
+++ b/pages/[type]/[id]/changelog.vue
@@ -239,9 +239,8 @@ function switchPage(page) {
     }
 
     .download {
-      position: absolute;
+      margin-left: auto;
       display: none;
-      right: 0;
 
       @media screen and (min-width: 800px) {
         display: flex;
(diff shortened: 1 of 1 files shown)


Nitrrine:
> For mobile, this still working ?
> 
> (There is no deployment preview yet)

Oh, I am forgot about this, let me check

Nitrrine:
Uh, download button hidden by default in mobile version 🤣

pages/[type]/[id]/changelog.vue
```diff
@@ -235,16 +239,14 @@ function switchPage(page) {
     }
 
     .download {
+      position: absolute;
```
Comment (Prospector) [inline pages/[type]/[id]/changelog.vue:242]: This is not the solution I'd use for this, because there will be overlapping on long titles. Instead, I'd use a `margin-left: auto` to push it to the far right.

Prospector (changes requested):

triphora (approved):

PR #1339, opened 2023-08-24
Limit project card's line height to 2 lines

Nitrrine:
Before:
Before
After:
After

### DIFF
--- a/components/ui/ProjectCard.vue
+++ b/components/ui/ProjectCard.vue
@@ -469,10 +469,16 @@ export default {
 }
 
 .description {
+  --max-lines: 2;
+  --line-height: 1.4;
+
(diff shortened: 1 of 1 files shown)


falseresync:
You should add an ellipsis and possibly an expand link

components/ui/ProjectCard.vue
```diff
@@ -469,10 +469,16 @@ export default {
 }
 
 .description {
+  --max-lines: 2;
```
falseresync, line 472: These should go to the :root tag with a prefix to keep consistency

Minenash:
I may be wrong, but wasn't there a limit to the length of the summary to prevent this? (That was added after this staging project was created)

PR #1340, opened 2023-08-24
Add translations for profile page

Mysterious-Dev:
All is in the title ^^.


Mysterious-Dev:
Ready for review ^^.

Mysterious-Dev:
My bad, I forgot some strings ^^'

Mysterious-Dev:
Now, Ready for review ^^

pages/user/[id].vue
```diff
@@ -104,30 +112,59 @@
             <div class="primary-stat">
               <DownloadIcon class="primary-stat__icon" aria-hidden="true" />
               <div class="primary-stat__text">
-                <span class="primary-stat__counter">{{ sumDownloads }}</span>
-                downloads
+                <IntlFormatted
+                  :message-id="messages.profileDownloadsStats"
+                  :values="{ count: sumDownloads }"
```
Comment (brawaru) [inline pages/user/[id].vue:117]: `formatCompactNumber` should be used here (retrieved using `useCompactNumber` composable). [Example](https://vintl.vercel.app/guide/extras/compact-numbers.html#using-compact-numbers)

pages/user/[id].vue
```diff
@@ -104,30 +112,59 @@
...
               </div>
             </div>
             <div class="primary-stat">
               <HeartIcon class="primary-stat__icon" aria-hidden="true" />
               <div class="primary-stat__text">
-                <span class="primary-stat__counter">{{ sumFollows }}</span>
-                followers of projects
+                <IntlFormatted
+                  :message-id="messages.profileProjectsFollowersStats"
+                  :values="{ count: sumFollows }"
```
Comment (brawaru) [inline pages/user/[id].vue:132]: `formatCompactNumber` should be used here as well.

pages/user/[id].vue
```diff
@@ -51,26 +51,34 @@
...
             </nuxt-link>
           </div>
           <template v-if="isEditing">
             <div class="inputs universal-labels">
-              <label for="user-username"><span class="label__title">Username</span></label>
+              <label for="user-username"
+                ><span class="label__title">{{
+                  formatMessage(messages.profileEditUsernameLabel)
+                }}</span></label
+              >
```
Comment (brawaru) [inline pages/user/[id].vue:75]:
```suggestion
              <label for="user-username">
                <span class="label__title">
                  {{ formatMessage(messages.profileEditUsernameLabel) }}
                </span>
              </label>
```

In Vue 3 new lines are automatically collapsed, so this Prettier fix only creates a mess with no benefit to it.

pages/user/[id].vue
```diff
@@ -51,26 +51,34 @@
...
               <input id="user-username" v-model="user.username" maxlength="39" type="text" />
-              <label for="user-bio"><span class="label__title">Bio</span></label>
+              <label for="user-bio"
+                ><span class="label__title">{{
+                  formatMessage(messages.profileEditBioLabel)
+                }}</span></label
+              >
```
Comment (brawaru) [inline pages/user/[id].vue:81]: Same here, Prettier made mess.

pages/user/[id].vue
```diff
@@ -259,6 +300,75 @@ const auth = await useAuth()
...
+    id: 'profile.stats.projects-followers',
+    defaultMessage: '<stat>{count}</stat> followers of projects',
+  },
+  profileJoinedAt: {
+    id: 'profile.joined-at',
+    defaultMessage: 'Joined {ago}',
+  },
+  profileJoinedAtTooltip: {
+    id: 'profile.joined-at.tooltip',
+    defaultMessage: '{date} at {time}',
```
Comment (brawaru) [inline pages/user/[id].vue:323]: You can format the dates right in the string, see [`date` type section of Message Syntax guide](https://formatjs.io/docs/core-concepts/icu-syntax#date-type).

brawaru (commented):
Some commentos 🪄

pages/user/[id].vue
Comment (Mysterious-Dev) [reply pages/user/[id].vue:117]: Why is it necessary to use `formatCompactNumbe` ?

Comment (brawaru) [reply pages/user/[id].vue:117]: It allows to display numbers in compact notation (e.g. 1,200 will be displayed as 1.2K in English and 1.200 in German). However, standard APIs aren't good at this, and that's [why `@vintl/compact-number` was created](https://github.com/vintl-dev/compact-number#prehistory).

It's already in Knossos, you just gotta get the formatter function using `useCompactNumber` composable, and then use it on `sumDownloads` here. 

You will also have to make sure that `count` plural doesn't use compound character `#` in plural cases, or otherwise it will be replaced with incorrect number. Instead, you should use the argument directly:

```
{count, plural, one {<stat>{count}</stat> download} other {<stat>{count}</stat> downloads}}
```

Comment (Mysterious-Dev) [reply pages/user/[id].vue:117]: So, if I understand, It's required for me to create a plugin before using `formatCompactNumber` function ?

Comment (brawaru) [reply pages/user/[id].vue:117]: No, it's already in Knossos — https://github.com/modrinth/knossos/blob/57964282ee52d5689dbb9a53a62d6681fdabb2a8/composables/compact-number.ts.

Just add the following to the `<script setup>`:
```js
const formatCompactNumber = useCompactNumber()
```

Comment (Mysterious-Dev) [reply pages/user/[id].vue:117]: Oh damn, I'm so stupid :/

pages/user/[id].vue
```diff
@@ -104,30 +112,59 @@
...
+                  formatMessage(messages.profileJoinedAt, { ago: formatRelativeTime(user.created) })
+                }}
               </span>
             </div>
             <hr class="card-divider" />
             <div class="stats-block__item secondary-stat">
               <UserIcon class="secondary-stat__icon" aria-hidden="true" />
-              <span class="secondary-stat__text"> User ID: <CopyCode :text="user.id" /> </span>
+              <span class="secondary-stat__text">
+                {{ formatMessage(messages.profileUserId) }} <CopyCode :text="user.id" />
```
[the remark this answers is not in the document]
Comment (Mysterious-Dev) [reply pages/user/[id].vue]: Changed ^^.

Mysterious-Dev:
Well I think that all requests have been processed. Thanks @brawaru for your help ^^.

brawaru (changes requested):

pages/user/[id].vue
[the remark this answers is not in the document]
Comment (Mysterious-Dev) [reply pages/user/[id].vue:1]: It's done. BTW, thanks again for your help ^^🙏

brawaru:
There are also messages that can be translatable in the code side of things, like error that the user is not found, that error has occurred, and meta description. Should be easy to implement. This error occurred message is common throughout Knossos, just like Edit, Save buttons.

Mysterious-Dev:
> There are also messages that can be translatable in the code side of things, like error that the user is not found, that error has occurred, and meta description. Should be easy to implement. This error occurred message is common throughout Knossos, just like Edit, Save buttons.

It's done ^^

pages/user/[id].vue
```diff
@@ -304,8 +414,10 @@ if (user.value.username !== route.params.id) {
 
 const metaDescription = ref(
   user.value.bio
-    ? `${user.value.bio} - Download ${user.value.username}'s projects on Modrinth`
-    : `Download ${user.value.username}'s projects on Modrinth`
+    ? `${user.value.bio} - ${formatMessage(messages.profileMetaDescription, {
+        username: user.value.username,
+      })}`
+    : `${formatMessage(messages.profileMetaDescription, { username: user.value.username })}`
```
Comment (brawaru) [inline pages/user/[id].vue:420]: I think since we're forming the text here, having two separate strings where one includes bio and other doesn't makes a lot more sense. I'm not sure other languages would use a dash (—) as separator, so yeah.

brawaru (changes requested):
Great job! Here's a few small things and it's ready.

Mysterious-Dev:
> Great job! Here's a few small things and it's ready.

All is done. I added a missing project type key (All) in user page and not in `i18n-project-types.ts` because it is used only once in the project

brawaru (changes requested):
Surely these are going to be the last requested changes from me? :Clueless:

GitHub doesn't allow comments on things that are outside of diff, so I'll list them here.

- https://github.com/Mysterious-Dev/knossos/blob/0a0cd27f294aaa6c320a77df5e57f952b11077f7/pages/user/%5Bid%5D.vue#L441-L458
  These should return `sum` without calling `$formatNumber` on it, otherwise all numbers above 1,000 will render as NaN.
- https://github.com/Mysterious-Dev/knossos/blob/0a0cd27f294aaa6c320a77df5e57f952b11077f7/pages/user/%5Bid%5D.vue#L507
  Missed translatable string. This one is a common one I believe.
- https://github.com/Mysterious-Dev/knossos/blob/0a0cd27f294aaa6c320a77df5e57f952b11077f7/pages/user/%5Bid%5D.vue#L199-L200
  Missed translatable strings. These are common, used in search as well.

Mysterious-Dev:
> Surely these are going to be the last requested changes from me? :Clueless:
> 
> GitHub doesn't allow comments on things that are outside of diff, so I'll list them here.
> 
>     * https://github.com/Mysterious-Dev/knossos/blob/0a0cd27f294aaa6c320a77df5e57f952b11077f7/pages/user/%5Bid%5D.vue#L441-L458
>       These should return `sum` without calling `$formatNumber` on it, otherwise all numbers above 1,000 will render as NaN.
> 
>     * https://github.com/Mysterious-Dev/knossos/blob/0a0cd27f294aaa6c320a77df5e57f952b11077f7/pages/user/%5Bid%5D.vue#L507
>       Missed translatable string. This one is a common one I believe.
> 
>     * https://github.com/Mysterious-Dev/knossos/blob/0a0cd27f294aaa6c320a77df5e57f952b11077f7/pages/user/%5Bid%5D.vue#L199-L200
>       Missed translatable strings. These are common, used in search as well.

All is done ^^.

brawaru (approved):
LGTM

triphora (approved):

PR #1344, opened 2023-08-26
Skip regenerate workflow on forks

brawaru:
also, shameless plug but check out #1200 kthx byee


Mysterious-Dev:
Love it! It's true that it's a real pain when you use the Github mobile application and it sends you a notification every 30 minutes when a workflow fails.

Mysterious-Dev (approved):
Love it! It's true that it's a real pain when you use the Github mobile application and it sends you a notification every 30 minutes when a workflow fails.

triphora (approved):

PR #1351, opened 2023-09-02
Fix dependency capitalization

triphora:
| Before: | After: |
|-|-|
| image | image |


Prospector (approved):

Prospector:
in general, any case of using CSS to capitalize should be replaced with localization eventually, but as a temp measure it works

PR #1356, opened 2023-09-05
Fix Ctrl+Click on version

DeVoidDev:
It's just been bothering me a lot while using the site, it doesn't matter if my code gets used or not (might be janky, I'm not a nuxt developer), but I just hope it can get fixed.

### DIFF
--- a/pages/[type]/[id]/versions.vue
+++ b/pages/[type]/[id]/versions.vue
@@ -41,13 +41,7 @@
         v-for="version in filteredVersions.slice((currentPage - 1) * 20, currentPage * 20)"
         :key="version.id"
         class="version-button button-transparent"
-        @click="
-          $router.push(
-            `/${project.project_type}/${
-              project.slug ? project.slug : project.id
-            }/version/${encodeURI(version.displayUrlEnding)}`
-          )
-        "
+        @click="clickVersion($event, version)"
       >
         <a
           v-tooltip="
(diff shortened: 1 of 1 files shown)


DeVoidDev:
Something like this would at least fix the ctrl clicking issue, but the real solution would be to change the design.
Because the real issue is that there are links inside links being used, which makes it difficult to use `nuxt-link`,
and the html spec discourages nested interactive content.

Minenash:
Ctrl-Click is equivelent to middle-clicking, could this also check for middle/right clicking (though right-clicking would be weird)

Another way was discussed in discord ([msg](https://discord.com/channels/734077874708938864/848329986577399818/1159634685740195922), but it's in #contributors):

*brawaru:*
> the best way you can do is use like spread link, but you gotta somehow make sure the z-indexes are correct so that button is not overlayed
the way it is done is that you find a link in a card, and add a ::after / ::before element that takes the entirety of the card
but then the problem will be that you won't be able to select any text, but it's already a problem so eh

> it's probably not that hard, just gotta make every card [relatively positioned](<https://developer.mozilla.org/en-US/docs/Web/CSS/position>) and [isolated](<https://developer.mozilla.org/en-US/docs/Web/CSS/isolation>), then stretch the version link, and change the z index of the button so that stretched link doesn't cover it

DeVoidDev:
I had attempted to do something like that, but because nested links technically aren't allowed by html spec, I just modified the existing elements
I'll try to see what I can do with the above info, but it might take a little while as I'm currently rather busy

PR #1360, opened 2023-09-09
Make .well-known/change-password redirect appropriately

### LINKED ISSUE #1317: The well-known location for changing a password does not redirect to the change password page
[A Well-Known URL for Changing Passwords](https://w3c.github.io/webappsec-change-password-url/) defines a well-known location for changing passwords. Despite Modrinth offering the ability for users to change their password, [`/.well-known/change-password`](https://modrinth.com/.well-known/change-password) does not redirect to the proper location.

As a result, password managers will offer fallback functionality for changing passwords, such as linking 'Change Password' buttons to the home page instead.

Mysterious-Dev:
All is in the title.

I think we can deploy this PR to test if password manager can use this ^^.

Close #1317

### DIFF
--- a/middleware/wellknown.global.ts
+++ b/middleware/wellknown.global.ts
@@ -1,5 +1,0 @@
-export default defineNuxtRouteMiddleware((to) => {
-  if (to.path === '/.well-known/change-password') {
-    return navigateTo('/settings/account')
-  }
-})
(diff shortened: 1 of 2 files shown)


brawaru (changes requested):
This is not a good way of doing that, since this will be a middleware running for every page. Instead, what you should do, is create a server route with redirect in `/server/routes/.well-known/change-password.ts`. It should be as simple as calling a `sendRedirect` function and returning.

Mysterious-Dev:
> This is not a good way of doing that, since this will be a middleware running for every page. Instead, what you should do, is create a server route with redirect in `/server/routes/.well-known/change-password.ts`. It should be as simple as calling a `sendRedirect` function and returning.

Done ^^.

brawaru (approved):
:shipit:

triphora (approved):

PR #1363, opened 2023-09-10
Update Discord invite link to use official

Mysterious-Dev:
All is in the title ^^.


Mysterious-Dev:
For config, it's needed because it was dead link

triphora:
The config also needs to be updated in https://github.com/modrinth/.github/blob/main/issue_templates/config.yml

Mysterious-Dev:
> The config also needs to be updated in https://github.com/modrinth/.github/blob/main/issue_templates/config.yml

It's already done ^^

triphora (approved):

PR #1386, opened 2023-10-01
Convert Avatar & Badge components to Composition API

Mysterious-Dev:
All is in the title ^^.

### DIFF
--- a/components/ui/Badge.vue
+++ b/components/ui/Badge.vue
@@ -34,7 +34,7 @@
   </span>
 </template>
 
-<script>
+<script setup>
 import ModrinthIcon from '~/assets/images/logo.svg'
 import ModeratorIcon from '~/assets/images/sidebar/admin.svg'
 import CreatorIcon from '~/assets/images/utils/box.svg'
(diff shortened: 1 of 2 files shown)


components/ui/Badge.vue
```diff
@@ -49,33 +49,16 @@ import LockIcon from '~/assets/images/utils/lock.svg'
 import CalendarIcon from '~/assets/images/utils/calendar.svg'
 import CloseIcon from '~/assets/images/utils/check-circle.svg'
```
triphora, line 50: Do you think you could migrate these to omorphia icons?

triphora (commented):
Looks great, just one thing

components/ui/Badge.vue
Mysterious-Dev: All icons ?

triphora: In this file yeah. Whenever I touch any file, I tend to also migrate stuff there to omorphia whenever I can

Mysterious-Dev: I can look to do that ^^

Mysterious-Dev: But it requires an omorphia PR (some icons are missing).

triphora: Don't worry about it then!

triphora (approved):

Mysterious-Dev:
> Don't worry about it then!

Okay, however I might one day make a PR for it ^^

PR #1387, opened 2023-10-01
Disable save & remove password button when no input password.

### LINKED ISSUE #1268: Save password button can be clicked even when there is no password set
### Describe the bug

Similar to #1266 

### Steps to reproduce

1. Go to account settings.
2. Click "Add password" button.
3. Don't type password.
4. Click on "Save password" button.

### Expected behavior

Save password button should be disabled until password is given to both fields.

### Additional context

_No response_

qtchaos:
Closes #1268

Disables the save/remove button in the change/add/remove password modal for better UX.

image
image

### DIFF
--- a/pages/settings/account.vue
+++ b/pages/settings/account.vue
@@ -92,7 +92,12 @@
             Cancel
           </button>
           <template v-if="removePasswordMode">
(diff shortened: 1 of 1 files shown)


Mysterious-Dev:
Close #1268

triphora (approved):

PR #1389, opened 2023-10-01
Fix license element deleting itself when pressing backspace.

### LINKED ISSUE #1324: When editing a project you can delete license element
### Describe the bug

When you are editing a project in the license tab while you are in the dropdown to select a license you can keep pressing backspace to remove text but if you do enough the whole element disappears.

### Steps to reproduce

1, Go to a project
2. Press settings
3. Head to the "License" tab
4. Click the license dropdown menu
5. Press backspace
6. Element disapears

### Expected behavior

Element to not disappear and dropdown textbox stay empty

### Additional context

_No response_

qtchaos:
Closes #1324

When using backspace on the license multiselect element, it will set license to null and break, fixed this using optional chaining and some other checks.

### DIFF
--- a/pages/[type]/[id]/settings/license.vue
+++ b/pages/[type]/[id]/settings/license.vue
@@ -44,12 +44,12 @@
             :close-on-select="true"
             :show-labels="false"
             :class="{
-              'known-error': license.short === '' && showKnownErrors,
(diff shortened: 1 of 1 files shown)


triphora (approved):
This was admittedly a very funny bug

PR #1393, opened 2023-10-03
Convert some components to Composition API & use omorphia icons

Mysterious-Dev:
All is in the title ^^.


Mysterious-Dev:
Should be done for omorphia with https://github.com/modrinth/omorphia/pull/100

PR #1394, opened 2023-10-03
Convert Chips component to Composition API & TypeScript

brawaru:
- Converted Chips to Composition API
- Uses generics for items now
- Capitalize property deprecated (it's not compatible w/ i18n)


brawaru:
Now that I think of it, this probably should go to Omorphia. Unfortunately Vue component libraries using SFCs are inherently not type safe, so these changes lack any usefulness.

PR #1397, opened 2023-10-05
Add Negative Filters

### LINKED ISSUE #516: Negative Search

This feature would instead, filter out stuff rather than filter in stuff.

 For instance, selecting Cursed will filter out Cursed mods. This could be done with a tristate button or a single button to instead do a negative search.

_No response_

### Additional context

_No response_

Minenash:
Supersedes #1013; closes #516.

Categories and loaders are self explanatory.

Though environments are a bit more complex:
- Except unsupported/unsupported, normal checking acts the same
  - if `Client` is checked, then mods with `client_side` set to `required` or `optional` will show
- Excludes are the inverse.
  - If `Client` is excluded, then mods with `client_side` set to `unsupported` will show. Allowing optional would make it equivalent to `Server` being checked, but also showing the pair server unsupported/client optional, which doesn't make sense here

Keeping this as a draft for now:
- Commented out code / a couple logs
- Not linted
- A large part of the toggle for Plugin loader is commented out. It didn't seem to work on the staging instance, but it should probably be decided / figured out before this is merged
- In #1013, one of the requests for change was to make the this type of checkbox it's own component instead of shoving it into the normal Checkbox. This has heavy code duplication, since the lines changed are minimal. Especially for the css section there's probably a way to not duplicate it, but I'm not knowledgeable enough in vue
- @Prospector had suggested having a button to toggle between include filters and exclude filters. (Though others disagree) The code changes to add this is minimal, it would still use the tri-state checkbox, but the logic in the toggle methods would be slightly changed

### DIFF
--- a/components/ui/search/SearchFilter.vue
+++ b/components/ui/search/SearchFilter.vue
@@ -1,7 +1,11 @@
 <template>
-  <Checkbox
+  <TriStateCheckbox
     class="filter"
-    :model-value="activeFilters.includes(facetName)"
+    :model-value="
+      activeFilters.includes(facetName) ? 1
+        : negativeFilters.includes(facetName) ? 2
+        : 0
+    "
     :description="displayName"
     @update:model-value="toggle()"
   >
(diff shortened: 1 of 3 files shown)


Prospector:
Will be implemented in a different way when we get around to redesigning the search page

PR #1398, opened 2023-10-05
Move extra loaders and change RP Resolutions to use OR

### LINKED ISSUE #953: Selecting multiple resource pack resolutions should be "or" instead of "and"

If I want to filter by all packs that are at most 32x32 (for example), I'd think that the correct way to do so would be to select all of the options up to and including 32x, yet doing so yields very few results. This is because it's actually checking for packs that have *all* of these resolutions.

As the title suggests, selecting multiple of these should combine them with an or instead of an and. For example, "8x or lower" + "16x" + "32x" should mean `"8x or lower" OR "16x" OR "32x"` (while the current is `"8x or lower" AND "16x" AND "32x"`).

It might be interesting for power users to be able to manually set the combination requirements in the search bar to allow for completely custom filters. This could look something like `((resolution:8x AND resolution:16x) OR resolution:32x) AND category:combat AND opensource:true` (or maybe using `&&`s and `||`s). If something like this is already implemented (wouldn't put it past you guys) then it would be helpful to make this feature a bit more overt as I could not find any information on it.

### Additional context

N/A

Minenash:
Right now, when the user clicks on `More` to show the rest of the loaders, the extra loaders are intermixed with the original. This is unitutive, and makes it harder to find what you're looking for and  to see what loaders are the ones behind the more.

This PR fixes that by putting all the loaders hidden by `More` at the bottom of the list

Edit:
Now also fixes #953, which changes the Resource Pack Resolution filters to be OR instead of AND

### DIFF
--- a/pages/search/[searchProjectType].vue
+++ b/pages/search/[searchProjectType].vue
@@ -42,12 +42,23 @@
               </h3>
 
               <SearchFilter
+                v-if="header === 'resolutions'"
+                v-for="category in categories.filter((x) => x.project_type === projectType.actual)"
+                :key="category.name"
+                :active-filters="orFacets"
+                :display-name="$formatCategory(category.name)"
+                :facet-name="`categories:'${encodeURIComponent(category.name)}'`"
+                :icon="null"
+                @toggle="toggleOrFacet"
+              />
+              <SearchFilter
+                v-else
                 v-for="category in categories.filter((x) => x.project_type === projectType.actual)"
                 :key="category.name"
                 :active-filters="facets"
                 :display-name="$formatCategory(category.name)"
                 :facet-name="`categories:'${encodeURIComponent(category.name)}'`"
-                :icon="header === 'resolutions' ? null : category.icon"
+                :icon="category.icon"
                 @toggle="toggleFacet"
               />
             </div>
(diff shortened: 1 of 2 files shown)


Prospector (changes requested):
I feel like the list of uncollapsed loaders should be a constant somewhere instead of repeating them

Minenash:
I also had that thought. I didn't realize loaders were listed in `composables/tag.js`. The list could be added there

Minenash:
Should it be a list major loaders, or hidden loaders

pages/search/[searchProjectType].vue
```diff
@@ -96,16 +89,8 @@
             <SearchFilter
               v-if="projectType.id === 'mod' && showAllLoaders"
               v-for="loader in tags.loaders.filter((x) => {
```
Comment (Minenash) [inline pages/search/[searchProjectType].vue:91]: Oh wait, this for can just be of hiddenModLoaders

Minenash:
My original plan was just to do  
`:active-filters="header === 'resolutions' ? orFacets : facets"`  
` @toggle="header === 'resolutions' ? toggleOrFacet : toggleFacet` " 

But the toggle one didn't work, so unless there's a way to do it, I ended up duplicating the search filter.

Using orFacets should be safe, since the only other thing that used them was loaders, but you don't have access to loaders on resource packs.

This does mean however that resolutions now use the g instead of f parameter, but that was going to be true anyways

Prospector (approved):

PR #1408, opened 2023-10-10
Add handling for locale files parsing errors

brawaru:
Added handling for errors that occur when message files get parsed.

If this happens, we'll try to parse message ignoring the tags (which is the most common mistake), which will look ugly, but at least be somewhat functional. If that fails too, we will just skip the string, in which case VIntl should use the default English string.

All such errors are also logged, which helps in figuring out which strings need fixing on Crowdin.


brawaru:
I'll draft it for now until I unpin FormatJS dependencies in `@vintl/compact-number` and `@vintl/how-ago`. While `as any` hack works, it's a hack, and there are still two of the same dependencies in lock file which is not great.

triphora (approved):

Geometrically (approved):

triphora:
Good once conflicts are resolved.

brawaru:
Rebased and resolved the conflicts.

PR #1409, opened 2023-10-11
Use omorphia icons for Badge component

Mysterious-Dev:
All is in the title ^^.

Related to #1386 and https://github.com/modrinth/knossos/pull/1386#pullrequestreview-1651922953.


Mysterious-Dev:
Should be part of omorphia.

PR #1413, opened 2023-10-16
Update to @vintl/how-ago 3.0.1

Mysterious-Dev:
All is in the title ^^.

This PR resolve the problem of rounding account age on the profile.

Thanks to @brawaru for the reactivity ^^.

Before :
image

After:
image

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
brawaru, line 1: You are using older version of pnpm, the one required for Knossos is 8.6.1 with lock file version 6.1. I'd recommend checking out [corepack](https://nodejs.org/api/corepack.html) to always use the correct version for a project you are working on.

brawaru (approved):
LGTM

triphora (approved):

PR #1424, opened 2023-10-26
Use FourOhFourNotFound component (from omorphia) for error page

Mysterious-Dev:
All is in the title ^^.


Mysterious-Dev:
Done in #1431

PR #1438, opened 2023-11-07
fix: import nuxt composables from #imports

danielroe:
This is a DX improvement when developing - we can avoid loading the entire barrel file at `#app` by using the new granular imports merged in https://github.com/nuxt/nuxt/pull/23951.


components/ui/modrinth-loading-indicator.ts
```diff
@@ -1,5 +1,5 @@
 import { computed, defineComponent, h, onBeforeUnmount, ref, watch } from 'vue'
-import { useNuxtApp } from '#app'
+import { useNuxtApp } from '#imports'
 import { startLoading, stopLoading } from '#imports'
```
triphora, line 3:
```suggestion
import { startLoading, stopLoading, useNuxtApp } from '#imports'
```

PR #1443, opened 2023-11-12
Add NeoForge to infer

ThatGravyBoat:
### What does this PR do?

It changes the inferring from assuming all mods with a `mods.toml` file is Minecraft Forge to checking if NeoForge is in the dependencies and on 1.20.2 and above only adding Minecraft Forge if `forge` is in the dependencies.

### DIFF
--- a/helpers/infer.js
+++ b/helpers/infer.js
@@ -123,6 +123,13 @@ export const inferVersionInfo = async function (rawFile, project, gameVersions) 
           .flat()
           .filter((dependency) => dependency.modId === 'minecraft')
 
+        if (mcDependencies.length > 0) {
+          gameVersions = getGameVersionsMatchingMavenRange(
+            mcDependencies[0].versionRange,
+            simplifiedGameVersions
+          )
+        }
+
         const hasNeoForge =
           Object.values(metadata.dependencies)
             .flat()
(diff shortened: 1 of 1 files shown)


Prospector:
Do all mods typically set forge or neoforge as dependencies like this? I'm slightly worried that older especially mods might not set forge like this, but maybe they do.

AzureDoom:
> Do all mods typically set forge or neoforge as dependencies like this? I'm slightly worried that older especially mods might not set forge like this, but maybe they do.

Yes, the Forge modid field has been consistent in the toml since as early 1.13.2 as per the MDKs that are listed on the site for the various versions (I've checked the latest for each version here from 1.20.1 to 1.13.2), this just needed for 1.20.2 where NeoForge has moved from the forge namespace.

ThatGravyBoat:
Yeah, the mods should be setting this and as far as I know, most mods do this, as this would be how you make sure that the API you use in a dev matches what version is used in production. Additionally, like AzureDoom said the forge MDK has been including this for years now so mods should have it, and creation tools like mc dev intellij plugin also add the dependency in the TOML when creating a dynamic template.

Geometrically (approved):

PR #1471, opened 2023-12-06
Analytics

Geometrically:
- [x] Clean up code/invoke SSR when possible
- [x] Fix styling in some places
- [ ] Make analytics routes only show up for those who have permissions (like on navigation)
- [ ] Add resolution selector for every type


Geometrically:
Superseded by #1483

PR #1483, opened 2023-12-13
New analytics

Geometrically:


composables/analytics.js
brawaru, line 1: Most of these appear to be utility functions rather than composables. Conventionally, utility functions belong to the [`utils/`](https://nuxt.com/docs/guide/directory-structure/utils) directory, which is functionally identical to `composables/`, but makes the distinction clearer.

composables/analytics.js
```diff
@@ -0,0 +1,330 @@
...
+
+// note: build step can miss unix import for some reason, so
+// we have to import it like this
+// eslint-disable-next-line import/no-named-as-default-member
+const { unix } = dayjs
+
+export const countryCodeToName = (code) => {
+  const regionNames = new Intl.DisplayNames(['en'], { type: 'region' })
+  return regionNames.of(code) || 'Unknown'
+}
```
brawaru, line 11: VIntl alternative which would use the user's language.

```js
export function useCountryNames(style = 'long') {
  const formattingOptions = { type: 'region', style }
  const { formats } = useVIntl()
  return function formatCountryName(code) {
    return formats.displayName(code, formattingOptions)
  }
}
```

<details><summary>TypeScript</summary>
<p>

```ts
export function useCountryNames(style: 'narrow' | 'short' | 'long' = 'long') {
  const formattingOptions = { type: 'region', style }
  const { formats } = useVIntl()
  return function formatCountryName(code: string) {
    return formats.displayName(code, formattingOptions)
  }
}
```

</p>
</details> 

Usage:
```ts
const formatCountryName = useCountryNames()
const taiwan = computed(() => formatCountryName('TW'))
```

[💠 Live demo](https://stackblitz.com/edit/vitejs-vite-jpk2bz?file=src%2FApp.vue)

composables/analytics.js
```diff
@@ -0,0 +1,330 @@
...
+
+export const countryCodeToFlag = (code) => {
+  if (countryCodeToName(code) === 'Unknown') {
```
brawaru, line 14: Formatted output is unreliable and should not be used for decision making. According to previous code, fake country code `XX` is used to denote unknown regions, so this check can be more straightforward.

```suggestion
  if (countryCodeToName(code) === 'XX') {
```

darling (approved):

PR #1486, opened 2023-12-15
Fix: "Set Password" button on profile page is not active

### LINKED ISSUE #1472: "Set Password" button on profile page is not active
### Describe the bug

image

When trying to set a password on an account that was previously only set up by GitHub login, the "Save password" button is not clickable. 

### Steps to reproduce

1. Go to the user settings page 
2. click "add password"
3. enter a password in both boxes
4. try to click "Save Password"

### Expected behavior

button is clickable, or an error message shows if the password is somehow invalid

### Additional context

Per emma's suggestion, removing the `disabled` attribute clientside allowed the button to be clicked, and the request succeded in updating my password -- this suggests it's a purely frontend issue?

Mysterious-Dev:
All is in the title ^^.

Close #1472.

### DIFF
--- a/pages/settings/account.vue
+++ b/pages/settings/account.vue
@@ -116,9 +116,9 @@
               type="button"
               class="iconified-button brand-button"
               :disabled="
-                  newPassword.length == 0 ||
-                  (auth.user.has_password && oldPassword.length == 0) ||
-                  newPassword !== confirmNewPassword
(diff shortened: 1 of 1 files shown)


Geometrically (approved):

Arthur-GYT:
@Geometrically @triphora For this fix, it is better to wait for a merge?

PR #1504, opened 2024-01-05
Add enter shortcut for 2FA & email management

Mysterious-Dev:
All is in the title ^^.

It's a QOL for account management.

### DIFF
--- a/pages/settings/account.vue
+++ b/pages/settings/account.vue
@@ -151,6 +151,7 @@
             maxlength="11"
             type="text"
             placeholder="Enter code..."
+            @keyup.enter="removeTwoFactor()"
           />
           <p v-if="twoFactorIncorrect" class="known-errors">The code entered is incorrect!</p>
           <div class="input-group push-right">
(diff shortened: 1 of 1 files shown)


pages/settings/account.vue
```diff
@@ -201,14 +203,15 @@
...
             <p v-if="twoFactorIncorrect" class="known-errors">The code entered is incorrect!</p>
           </template>
           <template v-if="twoFactorStep === 2">
             <p>
-              Download and save these back-up codes in a safe place. You can use these in-place of a
-              2FA code if you ever lose access to your device! You should protect these codes like
-              your password.
+              Download and save these backVerify-up codes in a safe place. You can use these
+              in-place of a 2FA code if you ever lose access to your device! You should protect
+              these codes like your password.
```
triphora, line 214:
```suggestion
              Download and save these back-up codes in a safe place. You can use these in-place of a
              2FA code if you ever lose access to your device! You should protect these codes like
              your password.
```

triphora (changes requested):

pages/settings/account.vue
Mysterious-Dev: Done ^^.

triphora (approved):

Prospector (approved):

PR #1517, opened 2024-01-07
Fix keypress listener on organisation member invite

rtm516:
When putting a username into the organisation manage members page and pressing enter an error occurs mentioning an invalid team.

The error:

Backend request causing the error:

This looks to just be a typo in the code which I have fixed by making the keypress listener run the same code as pressing the invite button.


rtm516:
Fixed in https://github.com/modrinth/knossos/commit/a037d24b0fee90ca91ad36a53be5367ec5ace0ae

PR #1560, opened 2024-01-10
Refactor scope labels for applications and pats

darling:


utils/auth/scopes.ts
```diff
@@ -1,44 +1,276 @@
...
-  COLLECTION_DELETE: BigInt(1) << BigInt(34),
-  ORGANIZATION_CREATE: BigInt(1) << BigInt(35),
-  ORGANIZATION_READ: BigInt(1) << BigInt(36),
-  ORGANIZATION_WRITE: BigInt(1) << BigInt(37),
-  ORGANIZATION_DELETE: BigInt(1) << BigInt(38),
-  SESSION_ACCESS: BigInt(1) << BigInt(39),
+const scopeDefinitions = [
+  {
+    name: 'USER_READ_EMAIL',
+    description: 'Read your email',
```
Prospector, line 4: Would be great to see localization support for these since they'll probably also need a slight refactor, but overall looks good

Prospector (approved):

PR #1562, opened 2024-01-11
Fix: Project banners aren't showing in collections

### LINKED ISSUE #1558: Project banners aren't showing in collections
### Describe the bug

Project banners aren't showing on selected collection page

Project banners are showing in search:
Снимок экрана (753)

Project banners aren't showing in collections (only gradient based on the icon):
Снимок экрана (754)

### Steps to reproduce

1. Go to Collections page
2. Go to any collection (if you don't have any, create one)
3. Add project(s) with custom banners
4. You'll see that project(s) in collection don't show banners, but only gradient based on the icon

### Expected behavior

Project banners are showing in collections

### Additional context

_No response_

Mysterious-Dev:
All is in the title ^^.

Close #1558.

### DIFF
--- a/pages/collection/[id].vue
+++ b/pages/collection/[id].vue
@@ -284,6 +284,12 @@
             :description="project.description"
             :downloads="project.downloads ? project.downloads.toString() : '0'"
             :follows="project.followers ? project.followers.toString() : '0'"
+            :featured-image="
+              project.gallery
+                .slice()
+                .sort((a, b) => b.featured - a.featured)
+                .map((x) => x.url)[0]
+            "
             :icon-url="project.icon_url"
             :name="project.title"
             :client-side="project.client_side"
--- a/pages/collection/[id].vue
+++ b/pages/collection/[id].vue
@@ -287,9 +287,6 @@
             :featured-image="
               project.gallery
                 .find((element) => element.featured === true)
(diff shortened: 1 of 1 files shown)


pages/collection/[id].vue
```diff
@@ -284,6 +284,12 @@
             :description="project.description"
             :downloads="project.downloads ? project.downloads.toString() : '0'"
             :follows="project.followers ? project.followers.toString() : '0'"
+            :featured-image="
```
Comment (Prospector) [inline pages/collection/[id].vue:287]: It should just filter by featured and show nothing if none are featured, it's a backend bug that the first non-featured one is shown in search.

Prospector (changes requested):

pages/collection/[id].vue
Comment (Mysterious-Dev) [reply pages/collection/[id].vue:287]: So I just need to use find() function?

Comment (Prospector) [reply pages/collection/[id].vue:287]: yeah, to get the first gallery image that is featured. if none, then show none

Comment (Mysterious-Dev) [reply pages/collection/[id].vue:287]: Done (I think) ^^.

Comment (Mysterious-Dev) [reply pages/collection/[id].vue]: Now that's fixed ^^.

Prospector (approved):

PR #1581, opened 2024-01-15
Add download stat on organization pages

misode:
Considerations:
* Should the "followers of projects" stat also be on organization pages?
* The stats on organization pages don't use `IntlFormatted`, unlike the ones on user pages

image

### DIFF
--- a/pages/organization/[id].vue
+++ b/pages/organization/[id].vue
@@ -92,7 +92,7 @@
                   <span class="primary-stat__counter">
                     {{ $formatNumber(projects?.length || 0) }}
                   </span>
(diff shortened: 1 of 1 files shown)


brawaru:
> The stats on organization pages don't use `IntlFormatted`, unlike the ones on user pages

If you are going for it already, replacing them with localisable messages would make this PR a lot better :P

I think we already have a message like that in profiles, so it has to be commonised (put in `utils/common-messages.ts`).

Prospector (approved):

Prospector:
well, can translate this later ig

PR #1583, opened 2024-01-16
[simple]: add title to leave project button

### LINKED ISSUE #1478: Explain why Leave project button is disabled

In the `members` section of a project's settings, there is a button labeled "Leave project" under the "Manage members" section. This button can sometimes be grayed out, (I assume if you are the project owner). However, there is no information displayed as to why you cannot leave the project.

Add some kind of indicator (like a tooltip) to that button that explains why it is grayed out.

_No response_

### Additional context

image

AdamRaichu:
Closes #1478

(Original pull deleted due to merge confilcts)

### DIFF
--- a/pages/[type]/[id]/settings/members.vue
+++ b/pages/[type]/[id]/settings/members.vue
@@ -48,6 +48,11 @@
         <button
           class="iconified-button danger-button"
           :disabled="props.currentMember?.is_owner"
+          :title="
+            props.currentMember?.is_owner
+              ? 'You cannot leave the project if you are the owner!'
+              : ''
+          "
           @click="leaveProject()"
         >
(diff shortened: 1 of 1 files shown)


Geometrically (changes requested):
This should use the v-tooltip API

Prospector:
> This should use the v-tooltip API

v-tooltip annoyingly does not work on disabled buttons

AdamRaichu:
> This should use the v-tooltip API

I don't know what that is, I'm not a react dev. Also, @Prospector says that won't work here.

@Prospector does that mean you think it should be merged as-is, or a different change should be made?

Prospector:
Will merge since it's at least a minor improvement

PR #1592, opened 2024-01-20
Fix: Unlisted projects are still visible on profile when they're not meant to be

### LINKED ISSUE #1545: Unlisted projects are still visible on profile when they're not meant to be.
I understand that they should be visible to you when logged in on your own profile, but logging out and refreshing, they're still visible:

image 

Project card:
image

Logged out, it's still on profile:
image

Logged in, expected to be visible, this is working:
image

Mysterious-Dev:
All is in the title ^^.

Close #1545


triphora:
This should be a backend change, not a frontend change.

PR #1616, opened 2024-01-28
Add translations for the default page layout

Prospector:


layouts/default.vue
```diff
@@ -463,118 +659,51 @@ function developerModeIncrement() {
...
+      body: {
+        url: window.location.href,
+      },
+    })
+      .then(() => {})
+      .catch(() => {})
+  })
+}
+function toggleMobileMenu() {
+  this.isMobileMenuOpen = !this.isMobileMenuOpen
```
brawaru, line 686: `this` refers to nothing here. There are a lot of mistakes like these in [the commit](https://github.com/modrinth/knossos/pull/1616/commits/d7a270abf690f358d991950184e6467dc8bc8a2e) to convert default layout to composition API that got chained to this PR.

```suggestion
  isMobileMenuOpen.value = !isMobileMenuOpen.value
```

brawaru (changes requested):
I have made a follow-up PR to this one @ https://github.com/modrinth/knossos/pull/1618. There are messages to commits that should hopefully explain rationale behind some of the changes that would've otherwise be review comments.

layouts/default.vue
Prospector: fixed

PR #1618, opened 2024-01-29
Follow-up fixes for #1616

brawaru:
- Re-organize default layout messages
- Fix navRoutes not being computed in default layout
- Fix untranslated Get Modrinth App buttons in default layout
- Make legal disclaimer in the default layout translatable

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -5,16 +5,16 @@
       class="email-nag"
     >
       <template v-if="auth.user.email">
-        <span>{{ formatMessage(messages.verifyEmailBannerTitle) }}</span>
+        <span>{{ formatMessage(verifyEmailBannerMessages.title) }}</span>
         <button class="btn" @click="resendVerifyEmail">
-          {{ formatMessage(messages.verifyEmailBannerButton) }}
+          {{ formatMessage(verifyEmailBannerMessages.action) }}
         </button>
       </template>
       <template v-else>
-        <span>{{ formatMessage(messages.addEmailBannerTitle) }}</span>
+        <span>{{ formatMessage(addEmailBannerMessages.title) }}</span>
         <nuxt-link class="btn" to="/settings/account">
           <SettingsIcon />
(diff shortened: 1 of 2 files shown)


Prospector:
i don't think the legal disclaimer should be translatable, because it's something required by the Minecraft Commercial use guidelines and they don't offer translated versions of that document. In general I don't think we could offer translated versions of legal verbiage that isn't written by a lawyer either.

layouts/default.vue
```diff
@@ -388,20 +392,20 @@
...
+          {{ formatMessage(messages.changeTheme) }}
         </button>
         <nuxt-link class="iconified-button raised-button" to="/settings">
           <SettingsIcon aria-hidden="true" />
           {{ formatMessage(commonMessages.settingsLabel) }}
         </nuxt-link>
       </div>
       <div class="not-affiliated-notice">
-        NOT AN OFFICIAL MINECRAFT SERVICE. NOT APPROVED BY OR ASSOCIATED WITH MOJANG OR MICROSOFT.
+        {{ formatMessage(footerMessages.legalDisclaimer) }}
```
Prospector, line 408: Don't think we should translate the message Microsoft requires to have since their CUG are only available in English and our crowdsourced translators are not lawyers

Prospector (approved):

Prospector (changes requested):
oops clicked wrong button, overall looks good other than that

layouts/default.vue
brawaru: per commit message:
> It wouldn't make much sense to leave it untranslated since it is meant for the end users too. It is also so small that it's unlikely to be mistranslated.

I don't see any surface for this to be mistranslated to justify dismissing this for ‘translators are not lawyers’ reason. This isn't a huge legal document. I don't see Mojang complaining about this either. If they do have a problem, they can always contact Modrinth.

Prospector: alright,we think it's ok

PR #1623, opened 2024-01-31
Fixed More Download Options button

ignissak:
Fixes https://github.com/modrinth/knossos/issues/1606
The issue was in bottom-transition div being too long.
image
The fix makes the buttons relative and pushing them to the front.

### DIFF
--- a/pages/app.vue
+++ b/pages/app.vue
@@ -1009,6 +1009,8 @@ useSeoMeta({
     margin: 0 auto;
     justify-content: center;
     mask-image: none;
+    position: relative;
+    z-index: 20;
 
     .outline-button {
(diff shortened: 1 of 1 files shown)


Prospector (approved):

pages/app.vue
```diff
@@ -1009,6 +1009,8 @@ useSeoMeta({
     margin: 0 auto;
     justify-content: center;
     mask-image: none;
+    position: relative;
+    z-index: 20;
```
Prospector, line 1013: 20 seems overkill, 1 should do, right? 20 may appear on top of other elements that are attempting to cover the screen.

Prospector (changes requested):

PR #1626, opened 2024-02-01
Add support for Swiss flag aspect ratio

sudolev:
Preview 
preview

Addresses #1604 (minor as it is).

Super small. Please feel free to close/request changes if there's something wrong like if I didn't follow some design/code convention you stick to.

### DIFF
--- a/components/ui/charts/ChartDisplay.vue
+++ b/components/ui/charts/ChartDisplay.vue
@@ -190,7 +190,13 @@
                 :key="name"
                 class="country-value"
               >
-                <div class="country-flag-container">
+                <div
+                  :class="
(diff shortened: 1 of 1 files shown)


brawaru (changes requested):
I'll let the maintainers judge the need for this change, but here are the changes I'd like to see.

Prospector:
To be honest, I don't think flags should be special cased like this. there's flags in all kinds of different aspect ratios, there's hardly a standard. I think it's fine if we make the icons for all the countries confirm to a set aspect ratio.

sudolev:
Completely fair call, super understandable! Still very glad I made this pr, I learned a lot (especially from your fixes @brawaru).

PR #1643, opened 2024-02-10
Migrate description page to Composition API & use omorphia components

Mysterious-Dev:
All is in the title ^^.

### DIFF
--- a/pages/[type]/[id]/settings/description.vue
+++ b/pages/[type]/[id]/settings/description.vue
@@ -16,7 +16,7 @@
       <MarkdownEditor
         v-model="description"
         :on-image-upload="onUploadHandler"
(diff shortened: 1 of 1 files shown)


pages/[type]/[id]/settings/description.vue
```diff
@@ -16,7 +16,7 @@
       <MarkdownEditor
         v-model="description"
         :on-image-upload="onUploadHandler"
-        :disabled="(currentMember.permissions & EDIT_BODY) !== EDIT_BODY"
+        :disabled="(props.currentMember.permissions & EDIT_BODY) !== EDIT_BODY"
```
Comment (Prospector) [inline pages/[type]/[id]/settings/description.vue:19]: shouldn't need `props.` in template

Prospector (changes requested):

PR #1645, opened 2024-02-13
Inconsistent min payout

ItsRemix:
Fix inconsistence of minimal payout between https://modrinth.com/dashboard and https://modrinth.com/dashboard/revenue


Prospector:
This changes the incorrect value, you can withdraw 0.01 with some gift cards

PR #1679, opened 2024-04-12
Add translation keys for modal creation

xinyihl:
Is there any specification for translating id？

Related to modrinth/code#1482

### DIFF
--- a/locales/en-US/index.json
+++ b/locales/en-US/index.json
@@ -276,7 +276,7 @@
     "message": "The visibility of your project after it has been approved."
   },
   "component.modal.modal-creation.visibility.placeholder": {
-    "message": "Choose visibility.."
+    "message": "Choose visibility..."
   },
   "component.modal.modal-creation.visibilitys.approved": {
     "message": "Public"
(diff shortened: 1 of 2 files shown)


locales/en-US/index.json
```diff
@@ -254,6 +254,45 @@
...
+    "message": "URL"
+  },
+  "component.modal.modal-creation.visibility.label": {
+    "message": "Visibility"
+  },
+  "component.modal.modal-creation.visibility.description": {
+    "message": "The visibility of your project after it has been approved."
+  },
+  "component.modal.modal-creation.visibility.placeholder": {
+    "message": "Choose visibility.."
```
notlin4, line 279: I think it is missing a dot.

xinyihl: I don't think so.
there were only two dots.

notlin4: Other placeholders all use three dots.

xinyihl: aright，i will fix it

PR #1684, opened 2024-04-16
Update pnpm to v9.3.0

brawaru:


brawaru:
Nothing significant, I feel. Just checked if it works locally and thought why not make a PR.

Geometrically (approved):

PR #1727, opened 2024-06-14
Fix page crash on orgs

Geometrically:


pages/[type]/[id].vue
```diff
@@ -1264,7 +1264,8 @@ const members = computed(() => {
         )
       : x.is_owner
   )
-  const rest = acceptedMembers.filter((x) => x.user.id !== owner.user.id) || []
+
+  const rest = acceptedMembers.filter((x) => !owner || x.user.id !== owner.user.id) || []
```
Comment (brawaru) [inline pages/[type]/[id].vue:1268]: I think it would've been better to just skip filtering altogether if owner is falsey :p 

```suggestion
  const rest = owner ? acceptedMembers.filter((x) => x.user.id !== owner.user.id) : acceptedMembers
```

PR #1730, opened 2024-06-15
Inject native router only in reactive scopes

brawaru:


brawaru:
This should fix issues like filter on versions page not working at all.

Prospector (approved):

PR #1731, opened 2024-06-15
add open in app button to mod & modpack project pages

evergales:
I noticed that theseus takes 'modrinth://' deeplinks for mods and modpacks

I thought it would be nice to add an 'open in app' button to those pages
because as far as I know there is currently no way (other than pasting code into console)
to use unlisted mods or modpacks in the launcher
and this could be a convenient way to make that a bit better
along with just being a nice qol feature

https://github.com/modrinth/knossos/assets/97823982/dd5405a2-5bce-467b-8799-8e45ef0da3a4

_I wasn't sure about making a thread for this in discord but can if needed!_

### DIFF
--- a/pages/[type]/[id].vue
+++ b/pages/[type]/[id].vue
@@ -329,7 +329,9 @@
                       hoverOnly: true,
                     },
                     { id: 'copy-id', action: () => copyId() },
-                    ...(project.project_type === 'modpack' || project.project_type === 'mod' ? [{ id: 'open-in-app', action: () => openInApp() }] : []),
+                    ...(project.project_type === 'modpack' || project.project_type === 'mod'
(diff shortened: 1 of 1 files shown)


Prospector:
We had planned on implementing it a different way, but I think doing it this way would be great for now!

One thing is, this should work for anything except for a mod that only supports plugin loaders I believe, since the Modrinth App supports everything but plugins.

evergales:
I think it should be correct now
but [#1227](https://github.com/modrinth/theseus/issues/1227#issue-2354874125) needs to be fixed
and I also think [UrlConfirmModal](https://github.com/modrinth/theseus/blob/7394fdc1624a36823151fd69218f47134008bd8e/theseus_gui/src/components/ui/URLConfirmModal.vue) is grabbing old project versions when installing

